Issue #16, opened 2026-09-14
[ARCH] Asynchronous GL/PBO capture path for Widgets and Quick targets


skawu:
Follow-up from SPIKE-01 (#3), recorded in `docs/capture-spike.md` (PR #15).

## Problem

SPIKE-01 showed that the synchronous public capture paths used for whole-window GL/Quick output can impose significant GUI/render synchronization cost:

| Path | measured on the SPIKE-01 host |
|---|---|
| `QOpenGLWidget::grabFramebuffer()` | 3.4-3.6 ms at 729x568, 55-61 ms on the first call |
| `QQuickWidget::grabFramebuffer()` | 3.05-13.30 ms at 729x568, unstable across identical runs |
| `QQuickWindow::grabWindow()` | 18.8 ms (Quick 2D) to 19.3 ms (Quick3D) at 960x600 |
| `QQuickWindow::grabWindow()` per frame on Direct3D 11 | the application's own event loop fell to ~1 fps while the call reported ~15 ms |

However, SPIKE-01 also measured the **public asynchronous** `QQuickItem::grabToImage()` path. Therefore this issue must not assume that all public Qt capture APIs block the GUI thread, nor jump directly to private/RHI/GL interception.

The unresolved question is whether a public asynchronous path can satisfy HyRemote's actual whole-window sharing requirements with acceptable composition fidelity, cadence, latency, buffer ownership and backpressure. Only if it cannot should lower-level GL/PBO/render-thread/version-specific mechanisms be introduced.

## Goal

Establish the best asynchronous/high-throughput capture path while preserving HyRemote's **Qt public API first** principle.

The work order is:

1. exhaust the viable public asynchronous path(s), especially `QQuickItem::grabToImage()` where applicable;
2. document exactly which HyRemote requirements they cannot satisfy, if any;
3. only then evaluate OpenGL PBO / render-thread / RHI or Qt-version-specific adapters for the remaining gap.

## Questions to answer

1. Can `QQuickItem::grabToImage()` capture the complete intended shared content for a `QQuickWindow` target, including Quick3D/custom-FBO content and the required sibling/overlay composition semantics?
2. What are its end-to-end latency, sustainable cadence, outstanding-request limit, memory behavior and backpressure characteristics?
3. Can multiple asynchronous captures be safely pipelined, and how are completed frames ordered/dropped?
[...]

Issue #17, opened 2026-09-14
[ARCH] GBM/DMA-BUF low-copy feasibility and RemoteFrame buffer ownership


skawu:
Follow-up from SPIKE-01 (#3), recorded in `docs/capture-spike.md` (PR #15).

## Problem

SPIKE-01 established one strong ownership fact and one provisional observation that must be kept separate:

| Hand-off form | status |
|---|---|
| borrowed pointer into producer-owned raw memory | **proven unsafe under reuse** in the host probe: every delayed consumer read observed rewritten content |
| shallow `QImage` copy | consumer content stayed stable because of implicit sharing / copy-on-write |

The original SPIKE draft attempted to quantify producer reallocations with `QImage::cacheKey()`. That metric is invalid for storage-allocation counting because `cacheKey()` identifies image contents and changes when the image is altered. Therefore any exact "N reallocations in M captures" claim is **withdrawn pending corrected storage-identity evidence** in PR #15.

The valid architecture conclusion remains: a zero/low-copy `RemoteFrame` cannot be represented as an unowned raw pointer. The producer/consumer contract needs explicit lifetime/recycle semantics.

## Goal

Decide whether a GBM/DMA-BUF low-copy path is technically viable from Qt on Embedded Linux/EGLFS, and define the buffer ownership model that any such path must implement.

## Questions to answer

1. Can Qt 6.8 on EGLFS export or expose the rendered surface as a DMA-BUF (or equivalent external buffer) using public APIs? If not, exactly which private/QPA/native hook is required and how is it isolated?
2. What synchronization is required between GPU producer and downstream consumer/encoder (fences, EGLImage/native fence, explicit sync, glFinish fallback, double/triple buffering)?
3. What ownership states must `RemoteFrame` carry? At minimum evaluate:
   - producer-owned/recycled buffer with explicit release;
   - ownership-transferred buffer returned to a pool;
   - immutable snapshot/copy;
   - external/native buffer with synchronization metadata.
4. How should slow-consumer backpressure work without blocking GUI/render production: bounded pool, drop-oldest/drop-newest, latest-frame semantics, or another policy?
5. Is a transport-neutral buffer pool the right core abstraction, and what API is needed without exposing GBM/DRM/RKMPP types in `hyremote-core`?
6. After PR #15 corrects its QImage storage probe, what does the actual detach/allocation behavior imply for the system-memory fallback pool size and reuse strategy?

## Non-goals

[...]

Issue #18, opened 2026-09-14
[SPIKE-01B] Validate accepted capture baselines on RK3588 / Qt 6.8.3 / EGLFS


skawu:
## Parent / evidence

Continuation of #3 after the accepted host evidence merged via PR #15 (`main@a25332bd0ad37ab8167a97aa882ec62c8645ae47`).

Governance mode: `transitional-explicit` until HyRemote#14 passes ADS machine acceptance.

This issue exists only because the original SPIKE-01 environment could not reach the required RK3588 + Qt 6.8.3 + EGLFS/OpenGL ES target. Desktop evidence must not be rerun as a substitute.

## Goal

Validate the already-selected **public-API correctness baselines** on the real embedded reference target and produce the missing evidence required to close #3.

## Required target

Record exact values, do not assume:

- RK3588 board/model and BSP/OS;
- Qt 6.8.3 Open Source build details;
- actual QPA (`eglfs`) and EGLFS device integration/plugin;
- EGL/GLES vendor, renderer and version;
- Qt Quick graphics API / render loop;
- display resolution / DPR;
- compiler/toolchain.

If the actual environment differs materially, report it rather than rewriting the target.

### E1 — QWidget baseline

Validate on EGLFS:

- `QWidget::render()` into caller-owned QImage storage;
- parent `QWidget::grab()` whole-window capture;
- child-widget/QPainter content fidelity;
- damage mapping logic from the merged host spike where applicable;
- capture timing avg/P95, process CPU, RSS trend and visible local-display impact;
- repeated start/stop.

### E2 — Qt Quick baseline

Validate:

- `QQuickWindow::grabWindow()` with Qt Quick 2D;
- local LCD remains correct during repeated capture;
- capture timing avg/P95 and sustainable cadence;
- GUI/event/render-thread impact;
- full-frame result and absence/presence of public damage geometry.

### E3 — content fidelity

Independently validate when the installed modules/application path support them:

- Quick3D;
- custom `QQuickFramebufferObject` / OpenGL-backed Quick content;
- QOpenGLWidget / QQuickWidget only if they are actually valid in this EGLFS configuration.

Unsupported combinations must be recorded as `unsupported` or `unverified`, never inferred from desktop behavior.

### E4 — ownership/lifetime sanity

Repeat only the minimum system-memory hand-off checks necessary to confirm that the merged `RemoteFrame` ownership requirements are not contradicted on the target. Do **not** implement DMA-BUF/GBM here; that belongs to #17.

[...]

Issue #5, opened 2026-09-14
[ARCH-01] Define HyRemote core API, frame lifecycle and threading model


skawu:
## Goal

Turn the architecture baseline into an implementation-ready core contract after the capture spikes have produced enough evidence.

## Scope

Define interfaces and ownership for:

- `Session` lifecycle;
- target registration/attachment;
- capture backend selection;
- `RemoteFrame` metadata and lifetime;
- damage regions;
- timestamps;
- backpressure/frame dropping;
- transport submission;
- input dispatch;
- optional hardware encoder integration.

## Constraints

The core must not expose or require:

- NeatVNC types;
- EGLFS private types;
- RKMPP types;
- GBM/DRM types unless represented through a generic external-buffer abstraction;
- Qt private/QPA headers.

Qt Widgets and Qt Quick must share the same core session model.

## Evidence-derived requirements

Accepted host-side evidence from #3 / PR #15 and #16 / PR #19 establishes these minimum constraints:

### Frame ownership

- a borrowed raw pointer into producer-reused storage is not a valid standalone frame contract;
- `RemoteFrame` needs explicit ownership/lifetime/release semantics;
- immutable snapshot/copy and pooled/recycled/external-buffer forms may coexist behind one transport-neutral model.

### Damage

- damage is optional;
- Widgets may provide normalized root-coordinate paint regions, but that is not automatically a transport-ready damage protocol;
- Qt Quick public capture may provide no damage geometry, so `damage unknown/full-frame` is a first-class state.

### Timestamp semantics

- **capture request time is not a valid `RemoteFrame` PTS by definition.**
- #16 evidence shows pipelined `QQuickItem::grabToImage()` requests can be fulfilled from a later rendered scene state; `tickInImage - tickAtRequest` is positive when the captured visual state advanced after the request.
- v0.1 must therefore stamp the frame at capture completion/ready time when no more accurate backend timestamp exists.
- a lower-level backend may provide a more accurate render/presentation timestamp, but the generic Core must not assume `request timestamp == frame PTS`.
- request time may be retained separately for diagnostics/latency accounting.

### Backpressure

- capture production must not block on a slow transport by default;
- completed frames need a bounded caller-owned queue;
- the default near-live policy should support drop-oldest / latest-frame-wins semantics;
- producer-throttle is a distinct optional policy, not the same behavior;
- queue capacity must have explicit semantics (queued frames vs the frame currently in service/transport ownership);
[...]

PR #19, opened 2026-09-14
spike: evaluate the public asynchronous capture path for #16

skawu:
## Linked issue

Relates to #16 — `[ARCH] Asynchronous GL/PBO capture path for Widgets and Quick targets`.

This is a **Draft** PR. It follows the issue's work order exactly: the **public
asynchronous path is evaluated first**, requirement by requirement, and no lower-level
GL/PBO, render-thread, RHI or version-specific mechanism is implemented. Scope discipline:
no DMA-BUF/GBM (#17), no VNC/RFB, no RKMPP/H.264, no QPA proxy, no Core public API, and no
private Qt API.

Two review rounds have been applied on this branch:

- **Round 1** (`5662534191`, `5662540092`): rebuilt the slow-consumer probe around one serial
  consumer with a strictly enforced queue bound, and narrowed the composition and visibility
  claims. Summary comment: `5662851185`.
- **Round 2** (`5663534836`): corrected the timestamp/freshness semantics - the measured
  quantity is a signed scene-state advance, not a lag, and the request time is not a valid
  content PTS. Summary comment: the latest comment on this PR. No measurement was re-run for
  that correction.

## Scope

- a throwaway harness under `spikes/async-capture/` that evaluates
  `QQuickItem::grabToImage()` against whole-window HyRemote requirements;
- composition, fidelity, latency, cadence, in-flight, ordering, request-to-image scene-state
  advance, memory, backpressure and failure-mode probes, each with a control;
- `docs/async-capture-spike.md` with the PASS/FAIL result per requirement, the metric
  definitions and the recommendation for the v0.1 asynchronous design;
- cross-references from `docs/capture-spike.md` and new host rows in
  `docs/compatibility.md`.

## Non-scope

- no GL/PBO, render-thread, RHI or `QQuickWindow::setGraphicsDevice` work (section 6.2 of the
  document states why it is not justified by this evidence);
- no DMA-BUF/GBM (#17), no hardware encoding, no transport, no public API freeze, no
  private Qt API.

## What was measured

Five scenes (`quick2d`, `quick3d`, `customfbo`, `quickwidget`, `openglwidget`), each QML
scene carrying two deterministic landmarks: an 8-bit **tick patch** that attributes a captured
image to a harness-visible scene state, and an **overlay sibling** parented to
`QQuickWindow::contentItem()` — the parent chain a same-scene QML overlay / `Popup.Item` uses,
deliberately not a descendant of the application's root item. Other popup modes
(`Popup.Window`, `Popup.Native`, Control-specific `Popup`/`ToolTip`) are **not** exercised.

Key results (full tables in `docs/async-capture-spike.md`):

| Finding | Evidence |
|---|---|
| The QML root item is **not** the whole window: a root-item grab misses the overlay (3.31 % differing pixels, 0.00 % outside the overlay rect vs the synchronous capture) | `*-quick2d-all.json`, `*-quick3d-all.json` |
| Grabbing `QQuickWindow::contentItem()` is pixel-identical to the synchronous whole-window capture for 2D and Quick3D (max channel diff 0, 0.00 % differing) | `*-quick2d-all.json`, `*-quick3d-all.json` |
| Custom `QQuickFramebufferObject` content is re-rendered per frame, not a stale texture (encoded counter 222 -> 250 over 29 distinct frames) | `host-windows-rhi-opengl-customfbo-all.json` |
| Pipelining does **not** produce more distinct frames: distinct ticks = requests / K, duplicates = requests - requests/K, and the **signed scene-state advance from request to captured image** averages exactly `(K-1)/2` — for K>1 the returned image is *newer* than the request, not older | `*-pipeline-k{1,2,4,8,16}.json` |
| The paced control gives one distinct frame per request with zero duplicates and zero request-to-image advance | `*-pipeline-paced-k4.json` |
| Completions are FIFO (0 out-of-order in every configuration) and the API drops nothing; duplicated content must be handled by the caller | all pipeline reports |
| Throughput scales with K (25.7 /s at K=1 to 191.6 /s at K=16) while latency (32.4 -> 80.2 ms) and GUI-thread jitter (1.23/12.54 -> 22.77/52.44 ms avg/p95) grow with it | `*-pipeline-k*.json` |
| Memory is bounded, not a per-request leak (+37.1 MiB at 240 requests vs +38.7 MiB at 1000; +0.9 MiB in the no-capture control) | `*-pipeline-k4-1000requests.json`, `*-pipeline-dryrun-control.json` |
| A 10 fps consumer needs a caller-owned bounded queue: unbounded, the backlog reached 55 frames / 123 MiB after 0.51 s and the consumer fell 5.8 s behind; with capacity 2 and latest-frame-wins the producer keeps 119.3 /s, retention stays at 6.6 MiB and 90.8 % of frames are dropped | `*-consumer-{unbounded-queue,drop-oldest-cap2,producer-throttle-cap2}.json` |
| The asynchronous path does **not** suffer from the Direct3D 11 stall: 119.1 captures/s at K=4 where the synchronous `grabWindow()` collapses to 1.1 iterations/s in the same session | `*-sync-baseline.json` |

## Timestamp semantics (input for ARCH-01)

The probe writes `tick` before each request and decodes the `tick` rendered into each captured
image, so it measures a **signed scene-state advance from the request to the captured image**
(`tick in image - tick at request`). A positive value means the image is **newer** than the
request, which is the normal outcome of pipelining: several pending requests are served by one
later render and all receive that newer state. This is **not** an age and not staleness, and
the earlier wording in this PR ("lags the request", "never newer than the request", "up to K-1
ticks old") was wrong and has been removed.

Consequences that are now stated in the document and in the recommendation:

- **The request timestamp is not a valid content PTS.** It does not identify the scene state
  contained in the returned image.
- The safe **v0.1 baseline is the completion / `ready()` timestamp**, which is also the moment
  the buffer enters the caller's ownership; the request time is kept only as a latency
  diagnostic for that request.
- A future lower-level backend that can observe the actual rendered or presented frame time may
  supply a more accurate PTS; the generic Core must accept a PTS from the backend rather than
  deriving one from the request.
- Consumer `staleFrames*` is a **request-sequence age** (how many later requests had completed
  before the frame was serviced), not visual content staleness. The consumer evidence is
  queue-depth/ownership bound, completion-to-service wall-clock age (21.8 ms avg, 170.2 ms max
  for drop-oldest), producer/consumer rates and delivered/dropped accounting.

## Requirement result

`PASS` on whole-window composition (with `contentItem()`), same-scene sibling/overlay-tree
content, Quick3D and custom-FBO fidelity, cadence, ordering, memory, public-API-only, and
backpressure with a caller-owned bounded queue plus a drop policy. Three gaps remain:

1. **R13** — the asynchronous path requires a **visible window**: a hidden window makes
   `grabToImage()` return a null result synchronously (no completion callback), while the
   synchronous path still captures it on this host. `minimized`/occluded were not tested.
2. **R14** — `QOpenGLWidget` has **no** public asynchronous capture API at all (the
   synchronous `grabFramebuffer()` measures 3.98 ms avg / p95 3.29 ms).
3. **R15** — `QQuickWidget` can capture its embedded Quick content asynchronously through
   `rootObject()`, but the surrounding widget composition still needs the synchronous parent
   `QWidget::grab()`.

`QQuickItem::grabToImage()` is therefore recommended as the **public-API asynchronous baseline
and v0.1 candidate** for native `QQuickWindow` targets, not as a claim about the final
high-performance production path (Qt documents the offscreen render plus GPU->CPU copy as
costly); the final performance choice stays evidence-driven by RK3588/EGLFS validation and by
#17. The gaps above are recorded and no lower-level path is implemented.

## Validation performed

| Check | Result |
|---|---|
| Build of the standalone async spike project (MSVC `/W4 /permissive-`, Release) | Clean, no warnings |
| Root build with `-DHYREMOTE_BUILD_SPIKES=ON` (both spikes) | Clean |
| `ctest --test-dir build/async-spike` | 7/7 passed |
| Composition control | Overlay present only in `contentItem()` and synchronous grabs, in every scene and on both backends |
| Fidelity control | Frozen scene with a pinned tick; repeat comparison is 0 differing pixels, so a reported difference cannot be time |
| Memory control | No-capture dry run in the same loop: +0.9 MiB vs +37.1 MiB with captures |
| Ordering control | Paced requests: 60 distinct ticks, 0 duplicates, 0 request-to-image advance |
| Consumer bounds | `maxQueueDepthObserved <= completedQueueCapacity` and `maxOwnedFramesObserved <= capacity + 1` hold in both bounded runs; `delivered + dropped == completed` in every configuration |
| Hidden-window probe | Executed in a child process so a stall cannot take the evidence run with it |

## Known limitations

- All numbers come from the SPIKE-01 Windows host. **RK3588/EGLFS/OpenGL ES remains
  unverified** and desktop results are not presented as a substitute.
- The custom-FBO scene renders continuously by design, so its fidelity comparison is
  tolerance-based (max channel diff 2) and the report includes both decoded counters. Its
  cadence and the `-all` synchronous baselines are the most load-sensitive figures in the set
  and are reported as ranges.
- The consumer is a **model** (one serial consumer, fixed service time, frame-count queue), not
  a real transport with variable service time, encoding or buffering.
- No probe measures the **visual age of the pixels** in a delivered frame; only queue depth,
  wall-clock completion-to-service age, rates, drop accounting and the separate
  request-to-image scene-state advance.
- Non-zero device pixel ratios, larger resolutions, multiple simultaneous Quick windows,
  multi-item request interleaving, `minimized`/occluded windows and non-`Popup.Item` popup
  modes were not tested.
- The harness itself had a result/connection reference cycle during development that made it
  measure its own retention; it was fixed and the whole evidence set was regenerated with the
  corrected probe. This is recorded in `docs/async-capture-spike.md` section 5.6.
- No `RemoteFrame` implementation exists; the ownership/lifetime, queue and PTS rules in the
  recommendation are requirements, not code.

## Result envelope

```
STATUS: completed (host-only; embedded half of the acceptance still open)
PR: <this pull request>
PUBLIC_ASYNC_VERDICT: QQuickItem::grabToImage() satisfies whole-window composition when the grabbed item is QQuickWindow::contentItem() (pixel-identical to the synchronous whole-window capture for 2D and Quick3D; root-item grabs miss same-scene sibling/overlay-tree content by exactly the overlay area), scales from 25.7 to 191.6 captures/s as the in-flight window grows from 1 to 16 while latency grows from ~32 to ~80 ms and GUI-thread jitter from 1.23/12.54 to 22.77/52.44 ms, returns FIFO completions with no drops but duplicated content for pipelined requests, and keeps memory bounded (+38.7 MiB for 1000 requests against +0.9 MiB in a no-capture control). On the same host and session it does not suffer from the Direct3D 11 stall that collapses the synchronous grabWindow() loop to ~1 iteration/s.
TIMESTAMP_SEMANTICS: the probe measures a SIGNED scene-state advance from request to captured image (tick in image - tick at request); a positive value means the image is newer than the request, which is the normal outcome of pipelining, and it is not an age or staleness measurement. Therefore the request timestamp is NOT a valid content PTS for RemoteFrame; the safe v0.1 baseline is the completion/ready() timestamp (which is also when the buffer enters the caller's ownership), the request time is kept only as a latency diagnostic, and a lower-level backend that can observe the real rendered/presented frame time may supply a more accurate PTS. The generic Core must accept a PTS from the backend rather than deriving it from the request.
CONSUMER_METRICS: consumer staleFrames is a request-sequence age (how many later requests had completed before service), not visual content staleness; the accepted freshness evidence is queue depth / owned-frame bound, completion-to-service wall-clock age (21.8 ms avg, 170.2 ms max with capacity 2 and latest-frame-wins), producer/consumer rates and delivered/dropped accounting.
BACKPRESSURE_POLICY: bounded completed queue with drop-oldest / latest-frame-wins so transport slowness never blocks capture/render - with a serial 10 fps consumer, capacity 2 keeps the producer at 119.3 /s, retains 6.6 MiB, drops 90.8 % of frames and serves frames at most 2 positions behind in the request sequence, whereas an unbounded queue reached 123 MiB of backlog and fell 5.8 s behind, and producer admission control drops nothing but pins capture to 9.8 /s.
GAPS_NOT_SOLVED: the asynchronous path requires a visible window (a hidden window yields a null result with no completion callback, while the synchronous path still captures it here; minimized and occluded untested); QOpenGLWidget has no public asynchronous API at all; a QQuickWidget can expose only its embedded Quick content asynchronously, not the surrounding widget composition.
LOWER_LEVEL_PATH: not implemented and not justified by this evidence - whole-window composition, Quick3D and custom-FBO fidelity are already met by the public path, and the two widget-target gaps have not been shown to be material against any target frame budget. Candidate follow-up scope if a budget appears: asynchronous QOpenGLWidget capture and asynchronous whole-window QQuickWidget composition, not "make the Quick path faster".
PRIVATE_API_REQUIRED: no.
RK3588_EGLFS_RESULTS: unavailable - no board or EGLFS environment is reachable from this workspace; nothing desktop-side is presented as embedded evidence.
OPEN_RISKS: embedded validation missing; the render loop is the real bottleneck so cadence depends on the scene's own frame cost; duplicated content is returned silently and a naive integration would waste bandwidth; the visible-window precondition may or may not matter per deployment; QQuickWidget/QOpenGLWidget whole-window sharing stays synchronous; the consumer is a model rather than a real transport; the visual age of delivered pixels was not measured; host variance makes the custom-FBO cadence and the all-mode synchronous baselines load-sensitive.
```


skawu:
## Owner Review — Round 1: FAIL / bounded correction required

总体评价：public-async-first 的执行顺序正确；`QQuickWindow::contentItem()->grabToImage()` 的合成/Quick3D/custom-FBO host 证据有价值；未越界进入 GL/PBO/RHI/private API 也符合项目原则。当前主要结论方向成立，但在把本 PR 作为 #5 Core backpressure/threading 事实源之前，存在一个证据级 BLOCKER，另有两处边界表述必须收窄。

### BLOCKER — slow-consumer / consumer-window probe does not implement the claimed consumer model

Evidence `host-windows-rhi-default-quick2d-pipeline-consumer-window2.json` reports:

- `consumerWindow = 2`
- `maxHeldObserved = 4`
- `peakRetainedBytes = 9,216,000`

At 960x600 ARGB32 this is exactly four retained frames, not two.

The code explains why: `consumerWindowFull` is computed once before the issue loop, then up to `maxInFlight` requests may still be issued; after completions, `held` can overshoot the configured consumer window. Therefore the current probe does not prove that a bounded window of 2 caps retained frames at 2.

There is a second modelling problem: `consumerDelayMs=100` does not model one serial 10 fps consumer. Every completed frame gets its own independent `releaseAfterNs = readyNs + 100 ms`, so multiple frames are effectively consumed/released in parallel. The current statement that the bounded run "settles at the consumer's rate" is therefore unsupported.

Required correction:

1. Define the intended slow-consumer model explicitly: one serial consumer/service clock, or a bounded ownership queue with a specified number of concurrent consumer slots.
2. If the intended model is a single 10 fps consumer, completion/release must be serialized (e.g. next service completion based on max(previousServiceEnd, readyNs) + serviceTime), not independently timed per frame.
3. Enforce the configured ownership/consumer window so `maxHeldObserved <= consumerWindow` (or rename/redefine the setting if it is only an admission threshold and document the overshoot bound).
4. Re-run bounded vs unbounded evidence and update the exact retained-memory, throughput and required-drop figures.
5. Keep the high-level conclusion only if the corrected evidence supports it: HyRemote needs bounded backpressure and a drop/coalesce policy so transport slowness never blocks capture/render.

This does **not** invalidate the composition/fidelity evidence or the need for caller-managed backpressure; it invalidates the current quantitative slow-consumer proof and the `window=2 -> consumer rate` claim.

### Documentation boundary corrections (non-architectural, but required before merge)

1. The spike directly tests an overlay-like sibling under `QQuickWindow::contentItem()`, not real Qt Quick Controls `Popup`/`ToolTip` modes. Qt 6.8 distinguishes `Popup.Item` (same-scene overlay), `Popup.Window` (separate top-level window) and `Popup.Native`. Restrict the claim to same-scene visual content under the `contentItem()` / overlay tree; actual popup modes remain compatibility cases to verify.
2. The failure probe calls `hide()`. Do not claim `minimized` until it is explicitly tested; record minimized as unverified.

### Accepted findings that do not need rework

- Grabbing the QML root item is insufficient for sibling/overlay-tree content; `contentItem()` is the correct whole-scene candidate for a native `QQuickWindow`.
- 2D and Quick3D fidelity against synchronous whole-window capture passed on the validated host; custom FBO passed within the documented tolerance and freshness probe.
- `grabToImage()` is asynchronous and avoids the D3D11 `grabWindow()` stall seen in SPIKE-01 on this host.
- Pipelining requires an explicit bounded in-flight policy; high K increases latency/staleness/jitter and returns duplicate scene states when multiple requests are issued before new renders.
- `QOpenGLWidget` has no equivalent public async API; `QQuickWidget` can only async-capture its embedded Quick content, not the surrounding QWidget composition.
- No lower-level GL/PBO/RHI/private path is justified for native `QQuickWindow` from the current host evidence.
- RK3588/EGLFS remains unverified.

After the slow-consumer probe is corrected and the two documentation boundaries are narrowed, I will perform Round-2 Review. PR remains Draft; #16 remains open.

skawu:
### Additional review note — positioning of `grabToImage()`

This is non-blocking for the correction round, but affects the architecture wording. Qt 6.8 documents `QQuickItem::grabToImage()` as an asynchronous correctness API that renders to an offscreen surface and copies GPU memory back to CPU, and explicitly warns that it can be costly for live preview. Therefore HyRemote should describe `contentItem()->grabToImage()` as the **public-API asynchronous baseline / v0.1 candidate**, not as the final high-performance production path. The final performance choice remains evidence-driven by RK3588/EGLFS validation and the low-copy investigation (#17). This does not justify implementing GL/PBO in #16 without a measured requirement gap.

skawu:
## Round-2 correction pushed — slow-consumer probe rebuilt, two claim boundaries narrowed

Governance mode: `transitional-explicit`. Correction instructions taken from `5662534191` and
`5662540092`. Scope: the one BLOCKER plus the two documentation boundaries. Nothing from
#17, no GL/PBO, no RHI or private Qt API, no VNC/RFB, no RKMPP, no QPA, no Core API. PR stays
**Draft**.

Commits on `spike/16-async-capture`:

- `661eae2` `test(spikes): replace the slow-consumer probe with a single serial consumer model`
- `ca43f0f` `test(spikes): regenerate async capture evidence with the corrected consumer model`
- `b79b571` `docs: correct the consumer evidence and narrow two claims`

---

## BLOCKER — the probe did not implement the consumer model it claimed

Both defects you identified were real, and both were in the probe rather than in the API:

1. **Parallel release.** `releaseAfterNs = readyNs + consumerDelayMs` was computed per frame,
   so several frames were "consumed" concurrently and the run did not model one 10 fps
   consumer.
2. **Admission computed once per loop slice.** `consumerWindowFull` was evaluated before the
   issue loop and up to `maxInFlight` requests could still be issued afterwards, so `held`
   could overshoot the configured window by up to K — which is exactly what
   `maxHeldObserved = 4` at `consumerWindow = 2` showed.

### Redesigned model

- **One serial consumer**: completed frames enter a single queue; the consumer takes one frame
  and is busy with it for `--consumer-service-ms` (100 ms here). Services are strictly
  serialized (`nextServiceStart = end of previous service`), so the sustained rate is
  `1000 / serviceMs` = 10 frames/s, which the measurements confirm (9.4-9.9 /s).
- **Capture in-flight K and the consumer queue are separate**: K bounds concurrent
  `grabToImage()` requests, the capacity bounds the completed queue.
- **The setting is renamed and redefined**: `--consumer-window` → `--completed-queue-capacity`.
  It bounds the **queue depth** strictly; the frame being serviced is owned in addition, so
  the total ownership bound is exactly `capacity + 1` frames. The bound is documented in the
  document, in the README and in every report (`consumerConfig.queueCapacityBounds`), and the
  old overshoot of up to K is gone.
- **Three producer policies, measured separately** so that the drop policy is not faked by
  stopping capture: `none` (unbounded queue), `drop-oldest` (latest-frame-wins) and
  `producer-throttle` (admission control, which also counts in-flight requests so the bound
  holds).

### Corrected evidence (`quick2d`, D3D11, `contentItem()`, K=4, one serial 10 fps consumer)

| Strategy | Requests | Producer /s | Consumer /s | Delivered | Dropped | Drop ratio | Max queue depth | Max owned frames | Peak retained | Frame age avg / p95 / max | Max stale frames | Load + drain |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| **none** (unbounded) | 60 | **116.9** | 9.7 | 60 | 0 | 0 % | **55** | 56 | **123 MiB** | 2848 / 5453 / 5786 ms | **55** | 0.51 s + 5.89 s |
| **drop-oldest**, capacity 2 | 240 | **119.3** | 9.9 | 22 | **218** | **90.8 %** | **2** | **3** | **6.6 MiB** | 21.8 / 70.2 / 170.2 ms | **2** | 2.01 s + 0.29 s |
| **producer-throttle**, capacity 2 | 60 | **9.8** | 9.5 | 60 | 0 | 0 % | **2** | **3** | **6.6 MiB** | 172.4 / 183.4 / 194.8 ms | **2** | 6.12 s + 0.28 s |

Files: `host-windows-rhi-default-quick2d-consumer-{unbounded-queue,drop-oldest-cap2,producer-throttle-cap2}.json`.
The two previous consumer files are deleted.

### Point-by-point response to the required corrections

1. **Model defined explicitly** - one serial consumer, service clock 100 ms/frame, documented
   in section 5.7 and in the report schema.
2. **Release serialized** - the next service cannot start before the previous one ends; there
   is no per-frame independent timing any more.
3. **Bound enforced** - `maxQueueDepthObserved <= completedQueueCapacity` now holds strictly
   (2 <= 2), and the setting was renamed to `--completed-queue-capacity` with the overshoot
   bound documented as exactly one in-service frame.
4. **Bounded vs unbounded re-run** and the retained-memory, throughput and required-drop
   figures updated: unbounded retains 123 MiB and falls 5.8 s behind after 0.51 s of capture;
   bounded retains 6.6 MiB and needs ~91 % drops (latest-frame-wins) or pins capture to the
   consumer rate (admission control).
5. **High-level conclusion kept, now supported** - the public API provides no queue, no
   capacity and no drop policy, so HyRemote must own them; the corrected evidence quantifies
   the cost of each policy instead of asserting a single one.

---

## Documentation boundary corrections

**1. Composition claim.** The validated claim is now **same-scene sibling content under
`QQuickWindow::contentItem()`** - the parent chain `Popup.Item` and `Overlay` use. Section 5.1
carries an explicit table: `Popup.Item` is marked as an *architectural inference* from the
measured sibling case and **not separately executed**; `Popup.Window`, `Popup.Native` and
Control-specific `Popup`/`ToolTip` usage are **unverified**. The scene/landmark descriptions in
section 4.1 and the harness README were narrowed the same way.

**2. Visibility claim.** Only `hide()` was exercised. Section 5.8 now states the measured case
(hidden: async request rejected with a null result while the synchronous `grabWindow()` still
returns an image here) and marks **`minimized` and occluded as unverified**, including in the
R13 row and the unverified list.

**3. Positioning.** `grabToImage()` is described throughout as the **public-API asynchronous
baseline / v0.1 candidate**, referencing Qt's own warning that the offscreen render plus
GPU->CPU copy can be costly for live preview, and stating that the final performance choice
remains evidence-driven by RK3588/EGLFS validation and #17. The compatibility rows say the
same. No GL/PBO work was added.

**Also recorded while re-running:** the `-all` runs re-measure the synchronous baseline in a
warmer process state and report higher figures for some scenes - `customfbo` 39.69 ms / 13.2 /s
/ GUI p95 78.21 ms versus 15.86 ms / 30.1 /s / 18.17 ms in its dedicated run. Section 5.4 now
states that the dedicated per-scene files are authoritative and that the duplicated
measurement is kept for transparency; the `customfbo` pipeline cadence varies the same way
(38.4 /s in the `-all` run versus 137 /s earlier) and is reported as a range.

---

## Verification performed

| Check | Result |
|---|---|
| Build of the async spike (MSVC `/W4 /permissive-`, Release) | Clean, no warnings |
| `ctest --test-dir build/async-spike` | 7/7 passed |
| Queue bound | `maxQueueDepthObserved = 2` with capacity 2 in both bounded runs; ownership bound 3 = capacity + 1 |
| Serial consumer rate | 9.4-9.9 frames/s against the 10 fps target in all three runs |
| Accounting | `delivered + dropped == completed` in every configuration, including the runs with no consumer modelled |
| Freshness control | Paced run still gives 60 distinct ticks, 0 duplicates, 0 lag |
| Memory control | No-capture dry run +0.9 MiB versus +38.7 MiB for 1000 captures in the same loop |
| Evidence schema | Every pipeline report carries `consumerConfig` and `consumer`; no report retains the withdrawn keys |
| Composition / fidelity / Quick3D / custom FBO | Re-run unchanged: 0 differing pixels for 2D and Quick3D, max channel diff 2 for the custom FBO scene |

## Unchanged / still open

- Composition, Quick3D and custom-FBO fidelity conclusions stand.
- RK3588 + Qt 6.8.3 + EGLFS/OpenGL ES remains **unverified**; nothing desktop-side is offered
  as embedded evidence.
- No lower-level GL/PBO/RHI/private path is justified by this evidence, and none was added.

## Result envelope

```
STATUS: completed (host-only correction; embedded acceptance still open)
PR: https://github.com/skawu/HyRemote/pull/19 (Draft, updated)
CONSUMER_MODEL: one serial consumer, one frame serviced at a time, 100 ms per frame (10 fps target, measured 9.4-9.9 /s); completed frames enter a single queue whose capacity bounds queue depth strictly, with the in-service frame owned in addition so the ownership bound is exactly capacity + 1.
BOUND_EVIDENCE: with capacity 2 the observed maximum queue depth is 2 and the maximum owned frames is 3 - the old overshoot of up to K frames is gone because admission now counts in-flight requests and the consumer no longer releases frames independently.
POLICY_COMPARISON: unbounded queue - 116.9 captures/s producer, 9.7/s consumer, backlog 55 frames (123 MiB) after 0.51 s, drain 5.89 s, delivered frames up to 55 frames / 5.8 s stale; bounded queue of 2 with latest-frame-wins - producer keeps 119.3/s, 90.8 % of frames dropped, 6.6 MiB retained, delivered frames at most 2 frames / 170 ms stale; bounded queue of 2 with producer admission control - no drops but capture pinned to 9.8/s with older delivered frames (172-195 ms).
CLAIM_BOUNDARIES: the composition evidence covers same-scene sibling content under QQuickWindow::contentItem() (the Popup.Item/Overlay parent chain) with Popup.Item an architectural inference and Popup.Window, Popup.Native and Control-specific Popup/ToolTip unverified; only hide() was tested, minimized and occluded are unverified; grabToImage() is positioned as the public-API asynchronous baseline / v0.1 candidate rather than the final high-performance production path.
PRIVATE_API_REQUIRED: no.
RK3588_EGLFS_RESULTS: unavailable - no board or EGLFS environment is reachable from this workspace; nothing desktop-side is presented as embedded evidence.
OPEN_RISKS: the consumer is a model with a fixed service time and a frame-count queue, not a real transport; the measurement with no consumer modelled reports full frames as delivered for accounting consistency; host variance makes the custom-FBO scene's cadence and the all-mode sync baselines load-sensitive, and both are reported as ranges.
```

skawu:
## Owner Review — Round 2: backpressure PASS / timestamp semantics correction required

The Round-1 BLOCKER is resolved. The corrected probe now models one serial consumer, separates capture in-flight from the completed-frame queue, enforces `maxQueueDepthObserved <= completedQueueCapacity`, and independently measures unbounded, drop-oldest/latest-frame-wins, and producer-throttle policies. The three new evidence files are internally consistent: capacity=2 yields max queue depth 2 and max owned frames 3; producer/consumer rates match the configured 100 ms serial service; and `delivered + dropped == completed` holds.

One Core-affecting semantic correction remains before merge:

### Timestamp / freshness terminology is currently reversed

The code defines:

`tickLag = tickInImage - tickAtRequest`

For K>1 the evidence is positive (`K=4`: avg +1.5, max +3). A positive value means the returned image contains a **later/newer scene state than existed when that request was issued**. Multiple pending requests are serviced by one later render and therefore receive the same newer visual state. It does **not** mean that the image is `K-1` frames older than its request.

Therefore the current statements such as:
- `content ... lags the request by up to K-1 ticks`;
- `content is never newer than the request`;
- R10 wording that the content is `up to K-1 ticks old`;

must be corrected.

This also gives an important ARCH-01 requirement:

> **request time is not a valid content PTS for `grabToImage()` under pipelining.**

A backend may keep request time for diagnostics, but a `RemoteFrame` presentation/content timestamp must be assigned at completion or supplied by a backend that can observe the actual rendered/presented frame time. The generic Core must not assume `request timestamp == frame PTS`.

### Consumer freshness metric boundary

The corrected slow-consumer result remains valid, but the current `staleFrames = producedSoFar - frame->index` metric is a request/completion-sequence age, not a measured visual-content age. Do not describe `max stale frames = 2` as proving the image pixels are exactly two rendered frames behind the latest scene.

For the accepted backpressure conclusion, use the evidence that is actually measured:
- queue depth / owned-frame bound;
- completion-to-service wall-clock frame age (`21.8 ms avg`, `170.2 ms max` for drop-oldest);
- producer/consumer rates;
- delivered/dropped accounting;
- capture pipeline's separate request-to-image scene-state delta.

It is fine to keep the existing numeric field for transparency, but rename or clearly define it as a **request-sequence age** rather than visual staleness.

### Required bounded correction

No measurement rerun is required unless the implementation/schema is renamed. Correct `docs/async-capture-spike.md`, README/PR/result-envelope wording, and any derived compatibility wording so that:

1. positive `tickInImage - tickAtRequest` is described as a scene-state **advance from request to captured image**;
2. request time is explicitly not used as RemoteFrame PTS;
3. completion/ready time is the safe v0.1 timestamp baseline, with future lower-level backends allowed to provide a more accurate render/presentation timestamp;
4. consumer `staleFrames` is not presented as visual-content staleness;
5. the accepted policy remains bounded completed queue + drop-oldest/latest-frame-wins so transport slowness cannot block capture/render.

No GL/PBO, RHI/private API, #17, VNC/RFB, RKMPP, QPA or Core implementation in this correction.

After this terminology/contract correction, #16 is ready to close and PR #19 can merge.

skawu:
## Timestamp / freshness semantics corrected — no measurement re-run

Governance mode: `transitional-explicit`. Correction instructions taken from `5663534836`.
Scope: terminology, the derived timestamp contract and the PR body. **No evidence was
regenerated and no report key was renamed**, so every file the Round-2 review validated is
byte-identical to what was reviewed - only the wording and the derivations that depended on the
reversed reading changed. Nothing from #17, no GL/PBO, no RHI or private Qt API, no VNC/RFB, no
RKMPP, no QPA, no Core API. PR stays **Draft**.

Commits:

- `dd0be55` `test(spikes): document the correct sign of the request-to-image scene-state advance`
- `3cb8413` `docs: correct the timestamp/freshness semantics and redefine the request-sequence age`

---

## 1. The sign convention is now stated as an advance

The corrected phrasing, now in `docs/async-capture-spike.md` section 5.5, in the harness header
and at the JSON emission site:

> `scene-state advance (request -> captured image) = tick in image - tick at request`
>
> A **positive** value means the captured image carries a scene state that is *newer* than the
> scene state at the moment the request was issued: the request was serviced by a render that
> happened later than the request itself. It is **not** an age or a staleness measurement, and a
> positive value must never be read as "the image is behind its request".

The reading that depended on the old sign has been rewritten: distinct ticks are still exactly
`requests / K` and duplicates `requests - requests/K`, but the average advance of `(K-1)/2` is
now explained as *all pending requests being served by one later shared render, so they receive
the same newer visual state* - a request does not hold a frozen copy of the scene as it was when
it was asked.

Removed everywhere (they now appear only in the notice that records their removal):

- `content lags the request by up to K-1 ticks`
- `content is never newer than the request`
- `up to K-1 ticks old` (including the R10 row)

The R10 row is redefined: the measured quantity is the signed advance, there is no measurement
of the content being stale relative to the request, and what the caller actually loses is the
correspondence between a request and a specific scene state.

## 2. Timestamp contract recorded for ARCH-01

Section 5.5 now ends with an explicit **Timestamp contract (input for ARCH-01)** block, and
recommendation 6 of section 7 was rewritten around it:

- **The request timestamp is not a valid content PTS for `grabToImage()` under pipelining** - it
  does not identify the scene state contained in the returned image.
- The safe **v0.1 baseline is the completion / `ready()` timestamp**, which is also the moment
  the buffer enters the caller's ownership (recommendation 7 keeps the ownership half of that
  statement).
- The request time is kept **only as a latency diagnostic** for that request.
- A future lower-level backend that can observe the actual rendered or presented frame time may
  supply a more accurate PTS, and **the generic Core must accept a PTS from the backend rather
  than deriving one from the request**.
- Recommendation 8 now derives transport frame selection from that PTS policy instead of from an
  assumed request/content relationship.

## 3. The consumer metric is a request-sequence age

Section 5.7 now separates the two quantities and forbids conflating them:

| Quantity | What it is | Status |
|---|---|---|
| `consumer.frameAge*Ms` | completion-to-service **wall-clock** age inside the serial consumer | measured elapsed time |
| `consumer.staleFrames*` | **request-sequence age** (`producedSoFar - frame index`): how many later requests had completed before this frame was serviced | ordering/backlog indicator, **not** visual content staleness |

The section states explicitly that "age in requests" is not "age of the picture" - a frame's
pixels correspond to whatever scene state its servicing render produced, and per section 5.5
that state can be *ahead* of the request time - and that **the visual age of the delivered
pixels is not measured by this spike**. The column is renamed in the table to
`Completion-to-service age` / `Max request-sequence age`, and the readings were reworded
accordingly (e.g. "at most 2 positions behind in the request sequence and 21.8 ms average
completion-to-service age", not "2 frames stale").

The accepted freshness evidence for drop-oldest is exactly the list you specified: **queue depth
/ owned-frame bound, completion-to-service wall-clock age (21.8 ms avg, 170.2 ms max),
producer/consumer rates, delivered/dropped accounting**, plus the **separate request-to-image
scene-state advance** from the capture pipeline.

The accepted policy is unchanged: **bounded completed queue + drop-oldest / latest-frame-wins so
that transport slowness cannot block capture/render**.

## 4. Where the wording changed

| Location | Change |
|---|---|
| `docs/async-capture-spike.md` | new Round-2 terminology correction notice; 4.1 landmark wording; 5.3 title and claim scoped to "re-rendered per frame, not request-relative age"; 5.5 definition, sign convention, table header, readings, timestamp contract; 5.7 definition block, column names, readings; R8/R10 rows; 6.1 wording; recommendation 1-9; risks; unverified list; section 10 report-key definition table |
| `spikes/async-capture/README.md` | pipeline probe description, metric-definition section with the sign convention and the timestamp rule, consumer metric definitions, tick-patch attribution wording |
| `spikes/async-capture/harness/async_probe.h` / `.cpp` | the field definitions now carry the sign convention, the PTS consequence and the request-sequence-age definition, including at the JSON emission sites (no key renamed) |
| PR body | corrected semantics, refreshed to the current evidence numbers, and the envelope below |

## Verification performed

| Check | Result |
|---|---|
| Rebuild of the async spike after the comment change | Clean, no warnings |
| `ctest --test-dir build/async-spike` | 7/7 passed |
| Evidence files touched | none - the reported schema, keys and numbers are unchanged |
| Search for the withdrawn phrasings | only the notice that records their removal remains |
| `docs/capture-spike.md`, `docs/compatibility.md` derived wording | checked, no timestamp/lag claim to correct (the SPIKE-01 "stale FBO texture" wording is a different, still-valid claim) |

## Unchanged / still open

- Backpressure result and policy: accepted, unchanged.
- Composition (same-scene sibling under `contentItem()`), Quick3D and custom-FBO fidelity:
  unchanged. `Popup.Window`/`Popup.Native`/Control popup modes and `minimized`/occluded remain
  unverified.
- RK3588/EGLFS remains unverified; nothing desktop-side is presented as embedded evidence.
- No lower-level path implemented or justified.

## Result envelope

```
STATUS: completed (host-only; embedded half of the acceptance still open)
PR: <this pull request> (Draft, updated)
TIMESTAMP_SEMANTICS: the probe measures a SIGNED scene-state advance from request to captured image (tick in image - tick at request); a positive value means the image is newer than the request, which is the normal outcome of pipelining - several pending requests are served by one later render - and it is not an age or staleness measurement. Therefore the request timestamp is NOT a valid content PTS for RemoteFrame; the safe v0.1 baseline is the completion/ready() timestamp (also the moment the buffer enters the caller's ownership), the request time is kept only as a latency diagnostic, and a lower-level backend that can observe the real rendered/presented frame time may supply a more accurate PTS. The generic Core must accept a PTS from the backend rather than deriving it from the request.
CONSUMER_METRICS: consumer staleFrames is a request-sequence age (how many later requests had completed before service), not visual content staleness; the accepted freshness evidence is queue depth / owned-frame bound, completion-to-service wall-clock age (21.8 ms avg, 170.2 ms max with capacity 2 and latest-frame-wins), producer/consumer rates and delivered/dropped accounting, with the request-to-image scene-state advance measured separately by the capture pipeline.
BACKPRESSURE_POLICY: bounded completed queue with drop-oldest / latest-frame-wins so transport slowness never blocks capture/render - with a serial 10 fps consumer, capacity 2 keeps the producer at 119.3 /s, retains 6.6 MiB, drops 90.8 % of frames and serves frames at most 2 positions behind in the request sequence, whereas an unbounded queue reached 123 MiB of backlog and fell 5.8 s behind, and producer admission control drops nothing but pins capture to 9.8 /s.
MEASUREMENT_RERUN: none - no schema key, formula or report was changed; the evidence files validated in Round 2 are byte-identical.
GAPS_NOT_SOLVED: the asynchronous path requires a visible window (a hidden window yields a null result with no completion callback, while the synchronous path still captures it here; minimized and occluded untested); QOpenGLWidget has no public asynchronous API at all; a QQuickWidget can expose only its embedded Quick content asynchronously, not the surrounding widget composition.
LOWER_LEVEL_PATH: not implemented and not justified by this evidence; RK3588/EGLFS validation plus #17 remain the inputs for any performance decision.
PRIVATE_API_REQUIRED: no.
RK3588_EGLFS_RESULTS: unavailable - no board or EGLFS environment is reachable from this workspace; nothing desktop-side is presented as embedded evidence.
OPEN_RISKS: the visual age of delivered pixels is not measured by any probe here; the consumer is a model rather than a real transport; host variance makes the custom-FBO cadence and the all-mode synchronous baselines load-sensitive; embedded validation missing.
```

skawu:
## Owner Review — Round 3 PASS

Round-2 backpressure correction and the final timestamp/freshness semantic correction are accepted.

Verified on head `3cb84135a1bedaa3efed0ec9225d71b83d280689`:

- `scene-state advance = tick in image - tick at request` is now defined consistently as a signed request-to-captured-image scene-state delta; positive means the captured image reflects a later/newer scene state.
- request timestamp is explicitly **not** used as `RemoteFrame` content PTS; completion/`ready()` time is the safe v0.1 fallback when the backend cannot provide a more accurate render/presentation timestamp.
- `consumer.staleFrames*` is explicitly defined only as request-sequence age / backlog ordering and is no longer presented as visual-content staleness.
- completion-to-service wall-clock age, queue depth/ownership bounds, producer/consumer rates, and delivered/dropped accounting are the accepted backpressure/freshness-adjacent evidence.
- Round-2 serial-consumer queue bounds remain valid; evidence files were intentionally not rewritten for the terminology-only correction.
- composition/visibility claims remain properly narrowed; RK3588/EGLFS stays unverified.

No further #16 exploration is required. `QQuickWindow::contentItem()->grabToImage()` is accepted as the public-API asynchronous **baseline / v0.1 candidate** for native visible QQuickWindow targets on the validated host, not as the final high-performance implementation claim.

Disposition: **PASS — ready to merge**. After merge, close #16 and move the active architecture work to #5. #18 and #17 remain independent evidence continuations.

Issue #21, opened 2026-09-14
[CORE-02] Implement hyremote-core Session, RemoteFrame and bounded dispatch pipeline


skawu:
## Parent architecture

Implements the accepted ARCH-01 baseline from #5 / PR #20 (`main@d1f7bff31c75f25789411fefd880038abf215619`).

Governance mode: `transitional-explicit` until HyRemote#14 passes ADS machine acceptance.

## Goal

Create the first production `hyremote-core` implementation that both #6 Widgets MVP and #7 Quick/EGLFS MVP can consume without duplicating Session/frame/backpressure logic.

This task implements the **semantic Core contract** only. It does not implement QWidget/QQuickWindow capture or NeatVNC.

## Required production modules

Create a Core library/target with implementation and tests for:

- `Session` lifecycle/state machine;
- `RemoteFrame` value and storage lifetime;
- CPU storage baseline and generic `StorageExtension` boundary;
- frame geometry / alpha / transform metadata;
- PTS/request/completion timing model;
- damage `Unknown | FullFrame | Regions`;
- capture request scheduling / max-in-flight accounting;
- bounded completed-frame mailbox;
- `DropOldest / LatestFrameWins` default policy;
- explicit `ProducerThrottle` alternative;
- Core dispatch worker;
- CaptureSource / Transport / InputSink interfaces;
- capture/transport capability compatibility checks;
- deterministic stop/fault cleanup.

## Architecture constraints

Use the accepted contracts in:

- `docs/adr/0001-core-boundaries.md`
- `docs/adr/0002-remoteframe-lifetime-timestamps.md`
- `docs/adr/0003-threading-backpressure.md`
- `docs/core-architecture.md`
- `docs/proposals/hyremote_core.hpp`

The proposal header is design input, not exact naming ABI. Implementation may refine names/signatures when tests expose a better source-level design, but it must not violate the invariants.

### Must not depend on

- Qt Widgets / Qt Quick / QML;
- Qt private/QPA APIs;
- NeatVNC / AML;
- GBM/DRM/DMA-BUF concrete types;
- RKMPP/V4L2/VA-API concrete types.

Prefer ordinary C++17 Core types. Internal QtCore use is allowed only if justified and does not leak GUI/private semantics into the public Core contract.

### C1 — frame lifetime

- consumer can retain a frame after producer callback returns;
- storage remains valid until final owner release;
- drop-oldest releases dropped frame ownership;
- a storage implementation with a custom recycler receives exactly one final release;
[...]

PR #20, opened 2026-09-14
docs(arch): define HyRemote Core contract (#5)

skawu:
## Linked issue

Closes #5 when accepted.

Governance mode: `transitional-explicit`.

## Goal

Turn the accepted host-side capture/backpressure evidence into an implementation-ready, protocol-neutral Core contract without freezing production ABI prematurely.

## Deliverables

- `docs/adr/0001-core-boundaries.md`
- `docs/adr/0002-remoteframe-lifetime-timestamps.md`
- `docs/adr/0003-threading-backpressure.md`
- `docs/core-architecture.md`
- `docs/proposals/hyremote_core.hpp` (design-only header proposal, not compiled/public ABI)
- updated `docs/architecture.md`

## Architecture decisions proposed

1. One Core `Session` model for Widgets and Quick; Qt target semantics live in adapters/facades.
2. `RemoteFrame` is immutable metadata plus shared owned `FrameStorage`; borrowed producer memory is not a valid standalone frame contract.
3. Frame storage may be CPU-readable or external; platform-specific external descriptors are exposed through a lifetime-owned `StorageExtension` boundary, while the concrete DMA-BUF/GBM descriptor stays deferred to #17.
4. Damage is explicit `Unknown | FullFrame | Regions`.
5. Request time is diagnostic only, not content PTS. Completion/`ready()` is the safe v0.1 fallback; a backend may supply render/presentation PTS.
6. Capture completion crosses a bounded Core mailbox before transport dispatch. Default near-live policy is `DropOldest / LatestFrameWins`; `ProducerThrottle` is separate and explicit.
7. Transport owns its runtime/event loop and downstream per-client backpressure; NeatVNC/AML types never enter Core.
8. Input is translated out of protocol-native form before the target adapter and marshalled to the target-required thread.
9. Encoding is optional downstream composition, not a mandatory Core pipeline stage.
10. `QQuickWindow::contentItem()->grabToImage()` remains a host-validated public-API v0.1 candidate only; #18 can tune EGLFS defaults/backend choice and #17 can refine external storage without changing Core invariants.

## Threading model

- Qt-affine capture initiation: target adapter / Qt GUI thread.
- Qt scene graph/render thread: Qt-owned; optimized adapters may hook it only under documented contracts.
- frame-ready callback: backend-defined thread, Core hand-off only.
- Core dispatch worker: isolates capture from transport.
- transport runtime: transport-owned (NeatVNC/AML isolated here).
[...]

Issue #9, opened 2026-09-14
[PERF-01][V1.0.0.0+x] Damage-aware/asynchronous capture on x86 with embedded follow-up


skawu:
## Scope authority update

Latest product ruling (#147/#157) changes the sequencing: **the portions that can be implemented and measured on Windows/Linux x86_64 belong in V1.0.0.0**. Embedded EGLFS/GBM/hardware-specific validation remains post-V1 in its own issues.

This supersedes the old dependency that PERF-01 could start only after the embedded Quick MVP (#7).

## V1 goal

Improve the existing desktop Widgets/Quick capture-to-transport path where deterministic x86 evidence proves a material benefit, without changing generic Core/application contracts.

### Required V1 work

- propagate useful damage information into incremental delivery in coordination with #144 B2;
- measure and, where supported, implement asynchronous graphics readback/capture paths that reduce render-thread stalls on x86;
- preserve portable public-API correctness baseline as fallback;
- keep frame pacing/drop policy bounded under client/network backpressure;
- measure capture timing and native UI responsiveness.

OpenGL/PBO is a candidate implementation, not the product API. If a reference x86 Quick backend is D3D11/RHI rather than OpenGL, use the applicable measurable path or explicitly record why PBO is non-applicable; do not force OpenGL merely to satisfy the issue title.

## Required measurements

On applicable Windows/Linux x86 reference configurations compare against the accepted correctness baseline:

- static/low-change UI;
- localized animated change;
- animated/full-change 2D UI;
- additional Quick3D/custom-GL case only where the reference configuration supports it.

Record at least:

- frame/capture time average + P95;
- bytes submitted/transferred where damage affects transport;
- process CPU;
- memory/buffer cost;
- delivered/update cadence;
- visible/native input/render responsiveness.

## Architecture constraints

- no Qt-private/platform type in stable application/Core contracts;
- portable baseline remains available;
- no unmeasured “accelerated” support claim;
- no embedded-only dependency in the V1 x86 build;
- optimized paths remain replaceable adapters/backends.

## Post-V1 / embedded continuation

Not required to close the V1 x86 part:

- #17 GBM/DMA-BUF/external-buffer ownership;
- #10 RK3588/RKMPP hardware encode;
- #18/#7 EGLFS target evidence/product line.

Desktop success must not be generalized to those targets.

## Acceptance

[...]

Issue #7, opened 2026-09-14
[MVP-QUICK] Deliver first Qt Quick/EGLFS end-to-end remote access path


skawu:
## Goal

Deliver the first end-to-end HyRemote path for a Qt Quick application running on the reference Embedded Linux/EGLFS stack while preserving local rendering and local input.

## Required flow

```text
QQuickWindow target
  -> Qt Quick target adapter
  -> v0.1 capture backend
  -> RemoteFrame
  -> first VNC/RFB transport
  -> standard VNC viewer

Viewer input
  -> transport callbacks
  -> Qt input adapter
  -> QQuickWindow/QML application
```

## Scope

- minimal Qt Quick example application;
- Embedded C++ integration API first;
- EGLFS/OpenGL ES reference validation;
- remote pointer and keyboard input;
- local display remains active;
- local touch/input remains usable;
- disconnect/reconnect without restarting the Qt application;
- baseline capture measurements documented.

### Required for MVP

- Qt Quick 2D controls/content;
- popup/overlay behavior where the selected capture path supports it;
- text input and focus;
- remote coordinate mapping;
- local/remote input coexistence.

### Compatibility evidence, not a blocker for first functional MVP

- Quick3D;
- custom OpenGL/FBO-backed Qt Quick content;
- QQuickFramebufferObject legacy/OpenGL path if available in the validation application.

These cases must be explicitly marked supported, experimental, or unsupported based on evidence from SPIKE-01.

## Non-goals

- production-grade high-motion performance;
- RKMPP/H.264;
- DMA-BUF zero-copy;
- zero-code QPA integration;
- Wayland migration.

## Deliverables

- `examples/quick-basic` or equivalent;
- Embedded Linux/EGLFS run instructions;
- implementation under shared HyRemote core abstractions;
- compatibility matrix entries;
- baseline performance measurements and known limitations.

## Acceptance

A Qt Quick/EGLFS example can add HyRemote through the Embedded C++ API, remain visible/interactive locally, and be remotely viewed and controlled from a standard viewer using the same HyRemote core model as the Widgets MVP.

## Dependencies

Depends on sufficient conclusions from #3, #4 and #5.

skawu:
## Dependency update

ARCH-01 (#5 / PR #20) is complete. This Quick/EGLFS MVP must consume the shared production Core implementation from #21 rather than creating its own Session/RemoteFrame/queue logic.

[...]

PR #22, opened 2026-09-15
feat(core): implement the hyremote-core Session, RemoteFrame and bounded dispatch pipeline (#21)

skawu:
## Linked issue

Implements [#21 [CORE-02] Implement hyremote-core Session, RemoteFrame and bounded dispatch pipeline](https://github.com/skawu/HyRemote/issues/21), based on the ARCH-01 baseline accepted through #5 / PR #20 (`main@d1f7bff`).

**Draft PR.** Production Core code only: no Widgets target adapter (#6), no Quick/EGLFS target adapter (#7), no NeatVNC/RFB transport (#9), no QML API, no QPA proxy, no DMA-BUF/GBM (#17), no RK3588/EGLFS validation (#18), no RKMPP/H.264, no encoder, and no binary ABI promise.

## Scope

- `hyremote-core` production library: Session lifecycle, `RemoteFrame` + storage lifetime, bounded completed-frame mailbox, drop policies, dispatch worker, `CaptureSource`/`Transport`/`InputSink` interfaces, capability check.
- Root CMake integration behind `HYREMOTE_BUILD_CORE` (ON by default), tests behind the existing `HYREMOTE_BUILD_TESTS`.
- 52 deterministic host-runnable test cases (C1-C7 of the issue) plus a dependency-boundary check.
- A short implementation-status section in `docs/core-architecture.md`.

## Layout

```text
core/
  CMakeLists.txt                     hyremote-core (+ HyRemote::Core alias), Threads only
  include/hyremote/core/
    types.hpp         ids, Size/Rect, PixelFormat, AlphaMode, FrameTransform
    storage.hpp       FrameStorage, StorageExtension, CpuFrameStorage (baseline + recycler hook)
    frame.hpp         FrameGeometry, Damage, FrameTiming/PtsSource, RemoteFrame, validation
    capabilities.hpp  CaptureCapabilities, FrameConsumerCapabilities, checkFrameCompatibility()
    capture_source.hpp CaptureRequest/Event, CaptureSource
    transport.hpp     Transport (bounded enqueueFrame contract), TransportEvent
    input.hpp         InputEvent placeholder schema, InputSink
    session.hpp       BackpressurePolicy, FrameQueueConfig, CaptureSchedule, SessionState,
                      SessionError, SessionStats, Session
    hyremote_core.hpp umbrella
  src/
    session.cpp, frame.cpp, storage.cpp, capabilities.cpp, types.cpp
    detail/mailbox.hpp|cpp   internal bounded mailbox (tested directly by C4)
  tests/
    support/{test_support.hpp,fakes.hpp}   tiny assertion helper + deterministic fakes
    test_{frame_lifetime,damage,timing,mailbox,transport_handoff,session_lifecycle,input_routing}.cpp
    check_dependencies.cmake               enforces ADR-0001 on includes + core/CMakeLists.txt
```

## Contract mapping

| #21 required module | Where |
|---|---|
| Session lifecycle/state machine | `Session`, `SessionState`; `Stopped -> Starting -> Running -> Stopping -> Stopped`, `Starting -> Faulted`, `Faulted -> Stopping -> Stopped` |
| `RemoteFrame` value + storage lifetime | `RemoteFrame`, `FrameStorage`, `CpuFrameStorage` |
| CPU storage baseline + generic `StorageExtension` boundary | `CpuFrameStorage` (single/multi-plane, release hook) + `FrameStorage::extension(domain/id)` |
| Frame geometry / alpha / transform metadata | `FrameGeometry` |
| PTS / request / completion timing model | `FrameTiming`, `PtsSource`, `normalizeFrameTiming()`, `requestToCompletionLatency()` |
| Damage `Unknown / FullFrame / Regions` | `Damage`, `Damage::hasRegions()` |
| Capture request scheduling / max-in-flight accounting | `Session` scheduler + `CaptureSchedule::maxInFlight`, `SessionStats::maxInFlightObserved` |
| Bounded completed-frame mailbox | `detail::Mailbox` (internal), `FrameQueueConfig::capacity` |
| `DropOldest` / `LatestFrameWins` default | `BackpressurePolicy::DropOldest`; the newest frame always survives |
| Explicit `ProducerThrottle` alternative | `BackpressurePolicy::ProducerThrottle`; admission counts in-flight requests |
| Core dispatch worker | `Session` dispatch thread: the only caller of `Transport::enqueueFrame()` |
| `CaptureSource` / `Transport` / `InputSink` interfaces | `capture_source.hpp`, `transport.hpp`, `input.hpp` |
| Capture/transport capability compatibility checks | `checkFrameCompatibility()`, enforced in `start()` before `Running` |
| Deterministic stop/fault cleanup | `Session::stop()`, `Session::abortStartup()` |

### Invariants and how they are held

1. **The capture callback never performs transport work.** `onFrameReady()` validates, stamps timing and pushes into the mailbox; it touches no transport code and holds no lock while doing so beyond the bounded bookkeeping.
2. **`enqueueFrame()` is treated as a bounded non-blocking post.** The dispatch worker calls it with no Core lock held; a transport that stalls or throws cannot block capture or terminate the process (the throw is counted and escalates to `Faulted`).
3. **Request time is never a PTS.** `normalizeFrameTiming()` accepts `Completion` (from `completionTime`) or an explicit `Render`/`Presentation` value, and *rejects* a frame whose only timestamp is `requestTime`; a declared Render/Presentation source without a value is rejected rather than relabelled.
4. **`FrameId` follows acceptance order.** The id is assigned inside the mailbox at push time, so it follows completion/acceptance order regardless of `requestId`.
5. **`DropOldest` is normal flow control.** It is counted in `SessionStats::framesDroppedByPolicy` and never raises an error or faults the Session.
6. **Queue capacity counts waiting frames; the dispatcher-owned frame is additional.** Peak Core-side ownership is `capacity + 1`, and `ProducerThrottle` admission counts in-flight requests so that bound cannot be exceeded even when several completions arrive at once (the accounting rule PR #19 round 2 established).
7. **The final storage release happens exactly once**, on the thread that drops the last reference; `CpuFrameStorage` runs its release hook from its destructor and the hook is documented as non-throwing.
8. **No forbidden dependency.** Core includes no Qt Widgets/Quick/QML/GUI/private, NeatVNC/AML, DRM/GBM/DMA-BUF or RKMPP/V4L2/VA-API header, and links no Qt target; this is enforced by a CTest case, not just by review.

## Deliberate refinements to the proposal header

The ARCH-01 header was design input, not exact API authority. The bounded changes (semantics preserved):

| Refinement | Reason |
|---|---|
| `Damage::regions(...)` factory renamed to `Damage::fromRegions(...)` | the proposal header cannot compile as written: the factory collides with the `Damage::regions` member |
| Added `SessionStats` + `Session::stats()` / `Session::config()` | tests and adapters both need to observe queue depth, drops, in-flight and dispatch counts; no observability was proposed |
| Added `validateSessionConfig()`, `sessionStateName()`, `sessionErrorCodeName()`, `pixelFormatName()`, `storageKindName()` | deterministic diagnostics for `SessionError` messages and for adapters |
| `Session::state()` is not `noexcept` | it takes the state lock |
| Added `CaptureSchedule::rejectedRequestRetryDelay` | a rejected request must not hot-spin the scheduler; the delay only affects the retry path |
| `CpuFrameStorage::mutablePlane()` + publication rule documented | the lifetime contract in ADR-0002 has no producer-side write accessor, which every Widgets/Quick adapter needs before publishing |
| The mailbox stays internal (`core/src/detail/`) and is tested directly | keeps the public surface to the contract; C4 still exercises the real unit |
| `InputEvent` is a minimal placeholder schema | ARCH-01 explicitly defers the normalized keyboard/pointer schema while freezing the threading/dispatch boundary |

## Acceptance evidence

Acceptance criteria of #21, in order:

1. **Core builds from the root project** — `HYREMOTE_BUILD_CORE=ON` (ON by default); `add_subdirectory(core)` in the root `CMakeLists.txt`; `HyRemote::Core` alias exported.
2. **Unit tests prove C1-C7** — 52 cases in 7 executables:

| Case | Tests | Cases |
|---|---|---|
| C1 frame lifetime | `hyremote-core-test-frame-lifetime` | 7 |
| C2 damage | `hyremote-core-test-damage` | 5 |
| C3 timestamps | `hyremote-core-test-timing` | 7 |
| C4 queue/backpressure | `hyremote-core-test-mailbox` | 8 |
| C5 transport hand-off | `hyremote-core-test-transport-handoff` | 7 |
| C6 lifecycle | `hyremote-core-test-session-lifecycle` | 13 |
| C7 input routing | `hyremote-core-test-input-routing` | 5 |

3. **No forbidden dependency** — `hyremote-core-test-dependency-boundary` scans every Core header/source include and `core/CMakeLists.txt`.
4. **Fake source + fake transport run end-to-end with bounded overload and deterministic stop** — `a_full_transport_runtime_queue_does_not_block_core`, `a_stalled_transport_cannot_block_the_capture_path`, `accepted_frames_are_drained_before_the_transport_stops`, `the_capture_in_flight_bound_is_respected`.
5. **#6 and #7 can depend on the same target without architecture changes** — the adapters only implement `CaptureSource`/`Transport`/`InputSink` and may only use the public headers; the mailbox is internal, no target-family type is mentioned in Core, and `SessionStats`/`SessionError` are the intended adapter-facing diagnostics.

### Exact commands and results (Windows 11 host, MSVC 19.44, Release)

```bash
# Core + tests only
cmake -S . -B build/core -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON
cmake --build build/core                 # clean, no warnings
ctest --test-dir build/core              # 8/8 targets, 52 cases, 0 failures
                                         # repeated 6x: 8/8 every time

# Full root project including both spikes (regression check)
cmake -S . -B build/all -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_PREFIX_PATH=C:/Qt/6.8.3/msvc2022_64 \
      -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON -DHYREMOTE_BUILD_SPIKES=ON
cmake --build build/all                  # clean
ctest --test-dir build/all               # 22/22 (14 spike smoke tests + 8 core targets)

# Default configuration (no explicit Core option)
cmake -S . -B build/default -G Ninja -DCMAKE_BUILD_TYPE=Release
# -> "Build core: ON", ctest 8/8
```

The Core tests need no GPU and no Qt GUI module: they do not link Qt at all, so they run on a bare host.

## Known limitations / not covered here

- **No capture-request timeout or fallback policy.** Core assumes a `CaptureSource` either completes a request or reports an event; a request that is silently dropped by a backend occupies an in-flight slot forever. Per-target timeout/fallback is adapter work (#6/#7).
- **The consumer-side freshness policy below Core is not implemented** (transport/client queues and their overflow policies are the transport's, per ADR-0003/#9).
- **No duplicate/coalescing detection** — ADR-0003 defers it and this PR only guarantees it is not inferred from request sequence.
- **No encoder, no external-storage concrete domain** (`StorageExtension` is the boundary only; #17 owns the descriptor).
- **Pacing is structural**: `targetFramesPerSecond` is implemented with a monotonic deadline, but no timing precision claim is made or tested.
- **RK3588/EGLFS is untouched**, so this PR proves host-side semantics only.
- Untracked local `logo/` files are not part of this PR.

## Result envelope

```text
STATUS: completed
PR: <this pull request>
CORE_TARGET: hyremote-core (CMake alias HyRemote::Core), C++17, links only Threads
TESTS: 52 deterministic cases in 7 executables + 1 dependency-boundary check = 8/8 CTest targets, 0 failures, stable over 6 consecutive runs; full root project 22/22 including the two spike suites
FRAME_LIFETIME: RemoteFrame carries shared_ptr<const FrameStorage> as the only lifetime anchor; a frame without storage, without geometry or without planes is rejected at acceptance; CpuFrameStorage owns its planes and runs an optional release hook exactly once from its destructor (verified including a cross-thread final release); DropOldest releases the dropped frame's ownership immediately; no naked pointer is ever published
TIMESTAMP_MODEL: FrameTiming{PTS, PtsSource(Completion|Render|Presentation), optional requestTime, optional completionTime}; Completion derives the PTS from completionTime, an explicit Render/Presentation value is preserved, a declared backend source without a value is rejected instead of relabelled, and a frame carrying only requestTime is rejected; request/completion diagnostics stay separately observable (requestToCompletionLatency); FrameId is assigned in Core acceptance order, not request order
BACKPRESSURE: bounded completed-frame mailbox, capacity counts waiting frames with the dispatch-worker-owned frame additional (peak ownership capacity + 1); DropOldest / latest-frame-wins is the default, counted in SessionStats and never an error; ProducerThrottle is explicit, drops nothing, and its admission counts in-flight requests so the bound holds even when completions arrive concurrently
THREADING: capture completion validates/stamps/enqueues only and never touches the transport; a single dispatch worker is the only Transport::enqueueFrame() caller and holds no Core lock across it; a transport that stalls (deterministically gated in tests) or throws cannot block capture or kill the process; Session::stop() stops scheduling, stops the capture source, closes and drains the mailbox, joins the workers, then stops the transport
FORBIDDEN_DEPENDENCIES: pass (no Qt Widgets/Quick/QML/GUI/private, NeatVNC/AML, DRM/GBM/DMA-BUF or RKMPP/V4L2/VA-API include; no Qt target in core/CMakeLists.txt) - enforced by a CTest case
ISSUES_FOUND_IN_ARCH_PROPOSAL: (1) the proposal header cannot compile as written - Damage::regions() collides with the Damage::regions member, fixed as fromRegions(); (2) no observability type was proposed for queue depth/drop/in-flight/dispatch accounting, needed by tests and adapters, added as SessionStats; (3) no producer-side write accessor or publication rule exists for storage, added CpuFrameStorage::mutablePlane() with the rule documented; (4) FrameStorage::mapRead() has no stated behaviour for storage without CPU mapping (returning nullopt is the implemented answer); (5) the proposal does not define what Core does with a completion that arrives after stop or with a backend that over-produces under ProducerThrottle - both are counted as anomalies rather than silently dispatched or dropped; (6) the free-standing InputEvent declaration cannot be implemented - a documented placeholder schema was needed
NEXT_READY: #6 (Widgets MVP) and #7 (Quick/EGLFS MVP) can consume the Core target; the Widgets and Quick adapters only need to implement CaptureSource/Transport/InputSink
```


skawu (commented):
## Owner Review — Round 1: FAIL / bounded correction

The overall Core structure is sound and the C1-C7 coverage is strong, but four correctness/lifecycle blockers must be fixed before this becomes the shared base for #6/#7.

### B1 — startup race can leave `Session` Running with a dead scheduler

`Session::start()` currently creates `schedulerThread` while `state == Starting`, then changes the state to `Running`. `runScheduler()` immediately returns when `state != Running`. A fast scheduler thread can therefore exit permanently before `start()` publishes `Running`; `start()` then returns true but no capture requests are ever issued.

Fix the startup handshake deterministically. Preferred shape: worker may be created during `Starting`, but waits on the condition variable until `Running | stop/fault`; after all worker creation succeeds, publish `Running` and notify. Do not rely on thread scheduling order. Add a deterministic regression test using a barrier/hook so the scheduler is forced to begin while the Session is still `Starting`, then prove it still issues requests after `Running` is published.

Also handle `std::thread` construction failure without leaving a half-started Session.

### B2 — `setCaptureSource()` / `setTransport()` can destroy live components

The public methods say components must be installed before `start()`, but the implementation does not enforce it. Calling either setter while Running can destroy the object while scheduler/dispatcher/callback code holds raw pointers to it, producing use-after-free.

Enforce mutation rules in code, not comments. Capture source and transport replacement must only succeed in `Stopped` state. Use an observable result (`bool`, error result, or equivalent) rather than silently replacing/ignoring. Add tests that attempt replacement while `Starting/Running/Faulted/Stopping` and prove the active component remains alive and unchanged. `InputSink` may be treated separately if runtime replacement is intentionally supported, because it is held through `shared_ptr`; document whichever contract is chosen.

### B3 — callback lifetime after stop/destruction is not defined

Callbacks installed into `CaptureSource` / `Transport` capture `&impl`. `Session::stop()` currently assumes adapters will stop safely, but neither interface states that `stop()` must quiesce callbacks before returning. A late frame/event/input callback after Session destruction can dereference freed `Impl` memory.

Freeze one safe rule and test it. Minimum acceptable contract: `CaptureSource::stop()` and `Transport::stop()` must not return until their installed callbacks can no longer be invoked, and fake adapters must model/verify that rule. A stronger Core-owned lifetime gate is also acceptable. The resulting contract must make `Session::~Session()` safe without relying on undocumented adapter behavior.

### B4 — plugin-boundary exceptions can escape `Session::start()` / input callback

Core catches exceptions from `requestFrame()` and `enqueueFrame()`, but not from `capabilities()`, `frameCapabilities()`, `CaptureSource::start()`, `Transport::start()`, or `InputSink::post()`. A throwing start currently escapes a bool-returning `Session::start()` and can leave `Starting` state / partially started components; a throwing input sink can escape through the transport runtime callback.

Make the boundary consistent: adapter exceptions must not escape Core-owned lifecycle/callback paths. Convert startup exceptions to a deterministic `SessionError` and clean up already-started components. Handle `InputSink::post()` failure without unwinding through the transport runtime (count/report/fault according to the chosen policy). Add deterministic tests for source-start throw, transport-start throw after source has started, capability-query throw, and input-sink throw.

### Keep unchanged

Do not expand into #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt target adapters. Existing frame ownership, timestamp, mailbox and dependency-boundary semantics remain accepted unless these fixes expose a direct contradiction.

Update the same PR #22. No new PR.

skawu:
## Round-1 correction pushed — B1–B4 fixed on the same branch

Governance mode: `transitional-explicit`. Correction instructions taken from review `5204202831`
and the Issue #21 owner disposition. Scope kept to the four blockers; no #6/#7, NeatVNC, DMA-BUF,
RKMPP or Qt adapter work, and the accepted frame ownership / timestamp / mailbox /
dependency-boundary semantics are unchanged. PR stays **Draft**, no new PR.

Commits:

- `df4f275` `fix(core): make the startup, mutation, callback and exception boundaries deterministic`
- `9644b61` `test(core): add regression tests for the lifecycle and boundary blockers`
- `5fc5bb2` `docs: record the Core lifecycle and boundary rules`

---

## B1 — startup race could leave a Running Session with a dead scheduler

**Root cause confirmed.** `runScheduler()` treated "not Running yet" as "exit", while the workers
were created during `Starting`; a scheduler that ran before `start()` published `Running` returned
for good.

**Fix.**

- Workers are created during `Starting` and perform an explicit **startup handshake**
  (`Impl::waitForRunning`): they wait on the condition variable until `Running` is published or the
  run fails/stops, then decide. `Running` is published *after* every worker was created
  successfully, followed by `notify_all()`.
- The dispatch worker uses `Impl::awaitStartup`, which waits for the handshake but never decides to
  exit there: it must still drain a closed mailbox if the run never reached `Running`.
- Start order is now capture source → mailbox + workers → transport → publish `Running`. Because
  no capture request is issued before `Running`, the transport is guaranteed to be started before
  any frame can be handed over.
- **`std::thread` construction failure** is caught: whichever worker exists is joined, the capture
  source is stopped, the mailbox is closed and the Session ends in `Faulted` with
  `SessionErrorCode::ComponentFailure` — never half-started. (This branch cannot be triggered
  deterministically, because `std::thread` construction failure needs resource exhaustion; it runs
  through the same `teardownRun()` path that the B4 startup-failure tests exercise.)

**Deterministic regression test.** `b1_workers_created_during_starting_survive_the_running_publication`
holds `Transport::start()` on a test gate, so the Session provably stays in `Starting` while both
workers exist (`SessionStats::workersStarted == 2`, a new diagnostic counter), asserts that no
capture request was issued yet, releases the gate and then requires the scheduler to keep working.

**Control experiment (recorded to prove the test is a real regression test).** With the handshake
temporarily replaced by the old racy logic, the case fails on exactly the intended assertion:

```
[ FAIL ] b1_workers_created_during_starting_survive_the_running_publication
         test_session_lifecycle.cpp:358: CHECK failed: waitFor([&] { return running.source->requestCalls() >= 1; })
```

The control edit was reverted immediately; the tree contains no trace of it.

## B2 — component replacement could destroy a live component

`setCaptureSource()` and `setTransport()` now return **`bool`** and only succeed while the Session
is `Stopped`. In every other state they refuse, leave the active component untouched, and release
the refused replacement object (documented in the header). `setInputSink()` stays runtime-
replaceable, which is now an explicit, tested contract rather than an accident: the sink is held
through `shared_ptr`, so swapping it cannot invalidate an in-flight callback, and a sink that is
already posted to may still be called once.

Tests: `b2_component_replacement_is_rejected_while_active`,
`b2_replacement_is_rejected_while_starting_and_while_faulted`,
`b2_replacement_is_rejected_while_stopping` (the last one calls the setter from inside the
adapter's `stop()` hook, where the state is provably `Stopping`). Liveness is proven with a
destructor counter on the fakes, not with a comment: the active source/transport report zero
destructions, the refused replacement reports exactly one, and the run keeps issuing requests.

## B3 — callback lifetime after stop/destruction

**Two layers, both now frozen.**

1. **Adapter rule (documented in `capture_source.hpp` / `transport.hpp`):** `stop()` must not
   return until the callbacks installed by `start()` can no longer be invoked. The fakes model
   this: `deliver()`, `deliverInput()` and `reportEvent()` refuse to call back once stopped, and
   only the explicit `force…` entry points can violate the rule.
2. **Core-owned lifetime gate:** the callbacks no longer capture the implementation by reference.
   They hold a shared `CallbackGate`; teardown closes the gate, waits for the callbacks already
   inside Core to finish, and only then invalidates the target. A late callback therefore finds
   `target == nullptr`, is ignored and counted (`SessionStats::callbacksIgnoredAfterStop`) without
   ever touching freed memory.

Tests: `conforming_adapters_stop_calling_back_once_stopped`,
`late_frame_and_event_callbacks_are_ignored`, `late_input_callbacks_are_ignored`, and
`late_callbacks_after_session_destruction_are_safe`, which keeps the adapters alive past
`~Session()` (through the new `LoaningCaptureSource`/`LoaningTransport` test adapters) and invokes
all four callback kinds afterwards — the gate makes it a counted no-op instead of a
use-after-free. As a consequence `Session::~Session()` is safe on its own contract, and `stop()`
is documented as "must not be called from inside an adapter callback".

## B4 — plugin-boundary exceptions

Every adapter entry point is now wrapped, and the policy is uniform and documented:

| Entry point | Policy |
|---|---|
| `capabilities()`, `frameCapabilities()` | caught → `SessionError{ComponentFailure, "capability query threw: …"}`, state stays `Stopped` (nothing was started) |
| `CaptureSource::start()` throw | caught → the component is stopped again (a throwing start may have left partial state) → `CaptureStartFailed`, `Faulted` |
| `Transport::start()` throw | caught → the throwing transport is stopped, the already-started capture source is stopped → `TransportStartFailed`, `Faulted` |
| `requestFrame()`, `enqueueFrame()` | counted (`captureRequestsRejected`, `transportEnqueueFailures`) and escalate to `Faulted` (existing behaviour) |
| `InputSink::post()` | caught inside the callback so it cannot unwind through the transport runtime; counted (`inputPostFailures`) and reported as a **recoverable** `SessionError` — remote input is not on the capture path, so frame delivery keeps working |

Tests: `b4_a_throwing_capture_start_faults_and_cleans_up`,
`b4_a_throwing_transport_start_stops_the_capture_source`,
`b4_a_throwing_capability_query_fails_before_starting` (both components),
`a_throwing_input_sink_is_reported_but_not_fatal`.

While correcting these paths, `stop()` also gained a single-threaded teardown guard, so concurrent
`stop()` calls cannot race on the same worker joins.

---

## Verification performed

```bash
cmake -S . -B build/core -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON
cmake --build build/core      # clean, no warnings
ctest --test-dir build/core   # 9/9 targets, 65 cases, 0 failures
                              # repeated 5x: 9/9 every time

cmake -S . -B build/all ... -DHYREMOTE_BUILD_SPIKES=ON     # full root project
cmake --build build/all       # clean
ctest --test-dir build/all    # 23/23 (14 spike smoke tests + 9 Core targets)
```

| Case family | Cases | Focus |
|---|---|---|
| C1 frame lifetime | 7 | unchanged |
| C2 damage | 5 | unchanged |
| C3 timestamps | 7 | unchanged |
| C4 queue/backpressure | 8 | unchanged |
| C5 transport hand-off | 7 | unchanged |
| C6 lifecycle | 20 | + B1 handshake regression, + 3 B2 mutation, + 3 B4 exception |
| C7 input routing | 5 | late-input case now asserts the gate path |
| B3/B4 callback + input boundary | 6 | new `test_callback_lifetime.cpp` |
| dependency boundary | 1 | unchanged |

## Kept unchanged

Frame ownership/lifetime, timestamp model, damage semantics, mailbox bounds and drop policies,
capability negotiation, and the "no forbidden dependency" rule are untouched, as are the spike
trees.

## Result envelope

```
STATUS: completed
PR: https://github.com/skawu/HyRemote/pull/22 (Draft, updated)
B1_STARTUP: workers are created during Starting and wait in an explicit handshake for the Running publication (or a stop/fault); Running is published only after all workers exist, followed by notify; no capture request is issued before Running and the transport is started before any hand-off; std::thread construction failure is caught and routed through the same deterministic teardown; regression test forces both workers to begin while Starting and then requires the scheduler to keep issuing, and it fails on exactly that assertion against a control build of the old logic.
B2_MUTATION: setCaptureSource()/setTransport() return bool and only succeed in Stopped; the active component is never destroyed (proven by destructor counters), the refused replacement is released, and refusal is verified while Starting, Running, Faulted and Stopping; the InputSink is explicitly runtime-replaceable through shared_ptr and covered by its own test.
B3_CALLBACK_LIFETIME: the adapter rule (stop() must not return before its callbacks can no longer fire) is documented on CaptureSource/Transport::stop() and modelled by the fakes; additionally every callback goes through a Core-owned gate that ignores and counts post-teardown callbacks and drains in-flight ones before invalidating the target, so ~Session() is safe even against a non-conforming adapter - verified for frame, capture-event, input and transport-event callbacks, including calls after the Session was destroyed.
B4_EXCEPTIONS: capabilities(), frameCapabilities(), CaptureSource::start(), Transport::start() and InputSink::post() are wrapped; startup exceptions become a deterministic SessionError with already-started components cleaned up; request/enqueue exceptions are counted and escalate to Faulted; input post failures are counted and reported as recoverable without unwinding through the transport runtime.
TESTS: 9 CTest targets, 65 deterministic cases, 0 failures, stable over 5 consecutive runs; full root project with both spike suites 23/23.
FORBIDDEN_DEPENDENCIES: pass (unchanged; still enforced by a CTest case).
NEW_OBSERVABILITY: SessionStats gains workersStarted, inputPostFailures and callbacksIgnoredAfterStop.
SCOPE: no #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt adapter work; frame ownership, timestamp, mailbox and dependency-boundary semantics unchanged.
```

skawu:
## Owner Review — Round 2: correction required (lifecycle teardown ownership only)

Round-1 B1–B4 are accepted. The correction is directionally sound and the new tests prove the intended fixes. Two remaining lifecycle races must be fixed before PR #22 can become the shared base for #6/#7.

### R2-B1 — concurrent `stop()` callers can both become the teardown owner

Current `stop()` checks `teardownStarted` under `impl.mutex`, unlocks/breaks when it is false, and only later `teardownRun()` reacquires the mutex and sets `teardownStarted = true`. Two callers can therefore both observe `false`, both break, and both enter `teardownRun()`. That can race on `source->stop()`, mailbox close/reset and the same worker `join()` operations.

The teardown claim must be acquired atomically while holding `impl.mutex`. Do not use a check-then-release-then-set sequence. Introduce one bounded ownership mechanism (`claimTeardownLocked`, token/state, or equivalent) shared by owner `stop()` and startup-failure teardown. Non-owners wait for the active teardown to finish and must never join/stop/reset the same run resources.

Add a deterministic regression test that releases two or more `stop()` callers simultaneously through a barrier and proves exactly one teardown owns the component stop/join sequence, all callers return, and final state is `Stopped`.

### R2-B2 — `stop()` during `Starting` can be undone by the still-running `start()` call

`stop()` is documented as safe from every state, but `start()` releases `impl.mutex` while calling component `start()` methods. A second thread can call `stop()` while the Session is `Starting`. Today teardown can complete to `Stopped`, after which the original `start()` call can resume, mark the source started, create workers/start the transport and finally publish `Running` again. The callback gate may already have been closed by the stop path.

Make lifecycle cancellation/serialization explicit. If `stop()` wins while `start()` is in progress, the in-progress start must not later publish `Running` or install a new run behind the completed stop. After every external/startup boundary (source start, worker creation, transport start), re-check the current run generation/state/cancellation before committing the next step, or use an equivalent lifecycle ownership design. Clean up any component that completed startup after cancellation.

Add deterministic tests using the existing start gates:
- stop while `CaptureSource::start()` is held in `Starting`;
- stop while `Transport::start()` is held in `Starting` with workers already created;
- in both cases, after the gate is released, `start()` returns false/cancelled, no request is issued, no later transition to `Running` occurs, components are stopped exactly once, and final state is `Stopped`.

### Keep unchanged

Do not reopen B1–B4, frame/timestamp/mailbox semantics, #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt adapters. This round is only lifecycle teardown ownership + start cancellation.

Update the same branch/PR #22. Keep Draft. Full evidence stays in GitHub; chat result remains one line only.

skawu:
## Round-2 correction pushed — teardown ownership and start cancellation

Scope is exactly the two lifecycle races from review `5672985281`: no reopening of B1–B4, no change
to frame/timestamp/mailbox semantics, and no #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt adapter work.
Same branch, same Draft PR.

Commits:

- `9643421` `fix(core): make teardown ownership atomic and cancel an in-progress start`
- `286b982` `test(core): add deterministic lifecycle-ownership and start-cancellation regression tests`
- `0a2ee8b` `docs: record the Core teardown-ownership and start-cancellation rules`

---

## R2-B1 — concurrent `stop()` callers could both become the teardown owner

**Root cause confirmed.** `stop()` checked `teardownStarted` under the mutex, released the mutex, and
only set the flag later inside `teardownRun()`. Two callers could both observe `false`, both leave the
critical section, and both enter `teardownRun()`, racing on `source->stop()`, the mailbox
close/reset and the same worker `join()` calls.

**Fix — one atomic claim shared by every teardown path.**

- `Impl::claimTeardownLocked()` takes the claim **while holding the mutex** (`teardownStarted = true`)
  and increments `runGeneration` in the same critical section. There is no
  check-then-release-then-set sequence anywhere on this path.
- `stop()` calls the blocking `Impl::claimTeardown()`, which waits for any in-flight teardown to
  finish and then claims atomically. Non-owners never touch the run: after their wait they find the
  run `Stopped`, release the claim (so further waiting callers can finish) and return.
- The startup-failure paths use `claimFailureTeardown()`, which claims under the same mutex and
  refuses when a concurrent `stop()` already owns the teardown.
- `teardownRun()` snapshots each run resource under the mutex (`sourceStarted`, `transportStarted`,
  `workersCreated`, mailbox, gate) and **clears the flags there**, so every component and worker is
  stopped/joined by exactly one side.
- `SessionStats::teardownsPerformed` makes the invariant observable: it is exactly 1 per run.

**Deterministic regression test.** `r2b1_concurrent_stop_callers_produce_exactly_one_teardown`
releases four `stop()` callers through a barrier and asserts that all four return, that exactly one
ordered teardown was performed (`teardownsPerformed == 1`), that each component was stopped once,
that nothing was destroyed, and that the Session is still usable afterwards (new source/transport,
fresh `start()`/`stop()` cycle).

**Control experiment.** With `claimTeardown()` temporarily restored to the old non-atomic shape
(check under the mutex, release, set the flag later), the case fails on exactly the intended
assertion:

```
[ FAIL ] r2b1_concurrent_stop_callers_produce_exactly_one_teardown
         test_session_lifecycle.cpp:351: CHECK_EQ failed: running.session->stats().teardownsPerformed == std::uint64_t{1}
```

## R2-B2 — `stop()` during `Starting` could be undone by the still-running `start()`

**Root cause confirmed.** `start()` released the mutex around the component `start()` calls, so a
second thread could complete a `stop()` and then the original `start()` would resume, mark the source
started, create workers, start the transport and publish `Running` again behind the completed stop
(with the callback gate already closed).

**Fix — explicit run generation, re-checked at every boundary.**

- `start()` claims a run generation under the mutex before touching any component
  (`generation = ++runGeneration`) and every teardown claim invalidates it.
- The generation is re-checked after **every** startup boundary and immediately before publishing
  `Running`, with the check and the commit in the same critical section:
  - callback-gate installation (nothing is installed when the run is already gone),
  - capture source start (commit of `sourceStarted`),
  - worker creation (mailbox + both threads are created under the mutex only when the run is still
    valid, so a teardown sees either "no workers" or "both threads assigned", never a half state it
    would join incorrectly),
  - transport start (checked before calling it, committed after it),
  - publication of `Running`.
- A cancelled `start()` never publishes anything: it stops exactly the component that finished
  starting **after** the cancellation (the teardown snapshot ran before that commit, so this is what
  keeps the stop count at one), reports `SessionErrorCode::StartCancelled` and returns false.
- Whatever `start()` committed *before* the claim is stopped by the teardown instead. Both directions
  are guarded by the same mutex, so each resource is cleaned up by exactly one side.

**Deterministic tests** (using the existing start gates):

- `r2b2_stop_while_the_capture_source_start_is_in_progress` — stop while `CaptureSource::start()` is
  held in `Starting`;
- `r2b2_stop_while_the_transport_start_is_in_progress` — stop while `Transport::start()` is held in
  `Starting`, with both workers already created and provably waiting in the handshake.

Both assert: after the gate is released `start()` returns **false**, the state is `Stopped`, no
capture request was issued (`requestCalls() == 0`, `framesDispatched == 0`), components were stopped
exactly once (`source` 1, `transport` 0/1 as appropriate), `teardownsPerformed == 1`, the error is
`StartCancelled`, and a later `start()`/`stop()` cycle works normally.

**Control experiment.** With the boundary validity check temporarily forced to `true` (the
pre-correction behaviour: no cancellation check at all), both cases fail on exactly the intended
assertion:

```
[ FAIL ] r2b2_stop_while_the_capture_source_start_is_in_progress
         test_session_lifecycle.cpp:395: CHECK failed: !started
[ FAIL ] r2b2_stop_while_the_transport_start_is_in_progress
         test_session_lifecycle.cpp:441: CHECK failed: !started
```

i.e. without the re-checks `start()` returns true and re-publishes `Running` behind the completed
stop. Both control edits were reverted; the tree contains no trace of them.

## API additions (additive, documented)

| Addition | Purpose |
|---|---|
| `SessionErrorCode::StartCancelled` | distinguishes "a concurrent stop() cancelled this start" from a component failure |
| `SessionStats::teardownsPerformed` | makes "exactly one ordered teardown per run" observable in tests and diagnostics |
| test-only `TestBarrier` + unbuffered harness output | deterministic release of concurrent callers; progress lines survive a crashing case |

The start-cancellation contract and the atomic teardown claim are documented in `session.hpp`
(`start()`/`stop()` comments) and in the implementation-status table of `docs/core-architecture.md`.

## Verification

```bash
cmake -S . -B build/core -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON
cmake --build build/core      # clean, no warnings
ctest --test-dir build/core   # 9/9 targets, 68 cases, 0 failures
                              # repeated 8x: 9/9 every time

cmake -S . -B build/all ... -DHYREMOTE_BUILD_SPIKES=ON
cmake --build build/all       # clean
ctest --test-dir build/all    # 23/23 (14 spike smoke tests + 9 Core targets)
```

| Case family | Cases |
|---|---|
| C1 frame lifetime / C2 damage / C3 timestamps / C4 mailbox | 7 / 5 / 7 / 8 |
| C5 transport hand-off | 7 |
| C6 lifecycle (includes the 3 new R2 cases) | 23 |
| C7 input routing | 5 |
| B3/B4 callback + input boundary | 6 |
| dependency boundary | 1 |

## Not changed

Frame ownership/lifetime, timestamp model, damage semantics, mailbox bounds and drop policies,
capability negotiation, the dependency boundary, and the previously accepted B1–B4 behaviour. Only
`core/src/session.cpp`, `core/include/hyremote/core/session.hpp`, the test support files,
`test_session_lifecycle.cpp` and one documentation table were touched.

## Result envelope

```
STATUS: completed
PR: https://github.com/skawu/HyRemote/pull/22 (Draft, updated)
R2_B1_TEARDOWN_OWNERSHIP: the teardown claim is now taken atomically while holding the Session mutex (claimTeardownLocked) and invalidates the run generation in the same critical section; stop() waits for an in-flight teardown and then claims atomically, non-owners find the run Stopped, release the claim and return without touching any component, worker or mailbox, and startup failures claim through the same mechanism; teardownRun snapshots each run resource under the mutex and clears its flag so every component/worker is stopped or joined exactly once; SessionStats::teardownsPerformed makes the invariant observable and is asserted to be exactly 1 for a run with four concurrent stop() callers released by a barrier - against a control build with the old non-atomic claim the case fails on exactly that assertion.
R2_B2_START_CANCELLATION: start() claims a run generation before touching any component and re-checks it after every startup boundary (gate installation, source start, worker creation, transport start) and immediately before publishing Running, with check and commit in the same critical section; a stop() that wins makes start() stop whatever it had started after the cancellation point, report SessionErrorCode::StartCancelled and return false without ever publishing Running, while everything committed before the claim is cleaned by the teardown instead - each component stopped exactly once; both required scenarios are covered deterministically with the existing start gates and fail on the !started assertion against a control build without the boundary checks.
TESTS: 9 CTest targets, 68 deterministic cases (65 previous + 3 new lifecycle cases), 0 failures, stable over 8 consecutive runs; full root project including both spike suites 23/23.
FORBIDDEN_DEPENDENCIES: pass (unchanged).
API_ADDITIONS: SessionErrorCode::StartCancelled, SessionStats::teardownsPerformed, test-only TestBarrier; start()/stop() contracts documented in session.hpp and docs/core-architecture.md.
KEPT_UNCHANGED: B1-B4 behaviour, frame ownership/lifetime, timestamp model, damage semantics, mailbox bounds and drop policies, capability negotiation, dependency boundary; no #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt adapter work.
```

skawu (commented):
## Owner Review — Round 3: FAIL / final bounded correction

R2-B1 teardown ownership and R2-B2 start cancellation are directionally correct and their required tests pass, but the correction introduced one mutex re-entry deadlock in three cancellation checks.

### R3-B1 — `start()` can self-deadlock while reporting cancellation

`Impl::finishCancelledStart()` acquires `impl.mutex` internally. `Session::start()` currently calls it from inside an already-held `std::lock_guard<std::mutex>` in three places:

1. callback-gate installation check;
2. pre-`Transport::start()` validity check;
3. final pre-`Running` publication check.

If a concurrent `stop()` invalidates the run before any of those checks, `start()` acquires `impl.mutex`, detects cancellation, and then recursively tries to acquire the same non-recursive mutex through `finishCancelledStart()`. The thread deadlocks.

The existing R2 tests do not cover these windows because they cancel while `CaptureSource::start()` or `Transport::start()` is externally gated and the cancellation is handled after the lock has been released.

### Required correction

Do not call any helper that acquires `impl.mutex` while `impl.mutex` is already held. For each affected boundary, detect cancellation under the lock, release the lock, then publish the cancellation diagnostic / return false. An equivalent locked-helper + unlocked-helper split is acceptable if the locking contract is explicit.

Add deterministic coverage for at least the two distinct shapes:

- stop wins immediately after `Starting` is published but before/while callback-gate installation is committed;
- stop wins after transport startup has committed but before `Running` publication.

Both must prove: no deadlock, `start()` returns false, no later `Running`, exactly one teardown, final `Stopped`, and a subsequent fresh start/stop still works.

Also inspect all `finishCancelledStart()` call sites and prove none can execute while `impl.mutex` is held.

### Keep unchanged

R1 B1-B4 and R2-B1/B2 semantics remain accepted. Do not change frame/timestamp/mailbox contracts, #6/#7, NeatVNC, DMA-BUF, RKMPP or Qt adapters. Same branch/PR #22; keep Draft. Full evidence stays in GitHub, chat response remains one line only.

skawu:
## Verification of the current PR head — build + full `ctest` (no code changed)

Independent verification of the state this PR currently points at. **No code was modified for this
verification**: the head commit was exported with `git archive` into a clean directory, so the result
cannot be influenced by any uncommitted local work.

### What was verified

| | |
|---|---|
| PR | #22 (Draft, OPEN) |
| Branch | `core/21-session-frame-pipeline` |
| Head commit | `0a2ee8b8ae7e3a87f71b0edd58671b00f82bbf05` (`docs: record the Core teardown-ownership and start-cancellation rules`) |
| Export | `git archive --format=zip -o head-0a2ee8b.zip origin/core/21-session-frame-pipeline` → clean tree (125 files) |

### Environment

- Windows 11 / x86_64, MSVC `14.44.35207` (x64) `cl.exe`, CMake + Ninja generator, `CMAKE_BUILD_TYPE=Release`
- Qt `6.8.3` `msvc2022_64`, `HYREMOTE_BUILD_CORE=ON`, `HYREMOTE_BUILD_TESTS=ON`, `HYREMOTE_BUILD_SPIKES=ON`, `HYREMOTE_BUILD_EXAMPLES=ON`

### Commands and results

```bash
cmake -S build/verify-head -B build/verify-head-out -G Ninja \
      -DCMAKE_BUILD_TYPE=Release -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DCMAKE_PREFIX_PATH=C:/Qt/6.8.3/msvc2022_64 \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON \
      -DHYREMOTE_BUILD_SPIKES=ON -DHYREMOTE_BUILD_EXAMPLES=ON
# Configure: done in 5.8 s. Only note: "Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)"
# (pre-existing and expected on this host; the same note appears for the documented builds).

cmake --build build/verify-head-out
# 0 errors, 0 warnings. 9.5 s (fresh build directory), second invocation: "ninja: no work to do".

ctest --test-dir build/verify-head-out
# 100% tests passed, 0 tests failed out of 23        Total Test time (real) = 39.43 sec
# 9 Core targets + 14 spike smoke tests, all Passed.
```

Core case counts in the head:

| Test executable | Cases |
|---|---|
| `hyremote-core-test-frame-lifetime` | 7 |
| `hyremote-core-test-damage` | 5 |
| `hyremote-core-test-timing` | 7 |
| `hyremote-core-test-mailbox` | 8 |
| `hyremote-core-test-transport-handoff` | 7 |
| `hyremote-core-test-session-lifecycle` | 23 |
| `hyremote-core-test-input-routing` | 5 |
| `hyremote-core-test-callback-lifetime` | 6 |
| **total deterministic cases** | **68** |

Stability: `ctest --test-dir build/verify-head-out -R hyremote-core` run 5× — `9/9` passed every time
(so the two Round-2 concurrency cases are not flaky on this host).

### Scope note (so the numbers are not over-read)

The verified head is the Round-2 state. It still contains the construct the Round-3 review flagged —
`Impl::finishCancelledStart()` acquiring `impl.mutex` internally (`core/src/session.cpp` line 534) and
the three `Session::start()` call sites that already hold that mutex — and it contains no `r3b1`
regression tests:

```
exported head: "static void finishCancelledStart" → L534      (present)
exported head: "CONTROL VARIANT"                  → none
exported head: "r3b1" in test_session_lifecycle   → none
```

So this verification confirms that the current head **builds clean and passes the full suite**; it does
not claim that the Round-3 finding is fixed. The Round-3 fix exists only in the working tree of the
verifying workspace and has not been committed or pushed.

For completeness, the local uncommitted work (not part of this head, no PR state attached) is
`core/src/session.cpp` + `core/tests/test_session_lifecycle.cpp` (+312/−69): it builds with 0 warnings
and passes `9/9` CTest targets / 26 lifecycle cases / 71 Core cases. A temporary local control patch used
to reproduce the R3-B1 failure mode was reverted before this verification and is not present in either
tree.

skawu:
## Owner disposition after verification comment 5673459606

The clean Windows/MSVC/Qt 6.8.3 verification is accepted as evidence for the current **Round-2 head** (`0a2ee8b`). It proves the checked-in head builds cleanly and passes 23/23 CTest targets / 68 Core cases on that environment.

It does **not** close Round-3: the verifier correctly confirmed that the checked-in head still contains the `finishCancelledStart()` mutex re-entry path and no R3 regression coverage. Therefore PR #22 remains Draft and unmerged until the R3 code correction is committed and reviewed.

Execution rule updated by Human Owner: work that can be completed through the current GitHub/tooling environment is performed directly by Owner/ChatGPT; Local Developer Agent is used only for genuinely local build/runtime/hardware evidence that cannot be produced here. No further implementation task is delegated for R3.

skawu:
## Owner execution — local-only publish step for Round 3

No new implementation design is requested.

The verified Round-3 fix exists only in the local verification workspace and is not present in the GitHub branch. This is the one remaining step that genuinely requires access to that local working tree.

Required local action only:

1. inspect the existing uncommitted Round-3 changes in `core/src/session.cpp` and `core/tests/test_session_lifecycle.cpp`;
2. confirm they are the same fix previously validated as 0 warnings / 9-of-9 Core CTest targets / 71 Core cases;
3. remove any temporary control-experiment residue if present;
4. commit those existing changes to `core/21-session-frame-pipeline`;
5. push to the same branch / PR #22;
6. rerun the existing clean build + full ctest once against the committed head and write the exact evidence here.

Do **not** redesign the fix, create a new PR, or expand into #39/#44/#43/#6/#28/#29/#27.

After the committed head appears on GitHub, Owner/ChatGPT resumes direct review/merge and subsequent product implementation.

skawu:
## Round-3 fix published and verified at the committed head

Executed the local-only publish step from review comment `5674192123`: the previously validated
Round-3 fix is now committed and pushed to the same branch / PR #22, and the committed head was
re-verified end to end. No redesign, no new PR, no other issue touched.

### 1–3. Local inspection before committing

| Check | Result |
|---|---|
| Uncommitted files | `core/src/session.cpp`, `core/tests/test_session_lifecycle.cpp` — 2 files, +312 / −69 |
| Fix shape | `Impl::finishCancelledStart()` (the helper that acquired `impl.mutex` internally) is gone; it is replaced by the locked-only `Impl::markStartCancelledLocked()` (session.cpp L538), whose contract is "caller holds `mutex`; this never acquires it" |
| Call sites | 9 `markStartCancelledLocked` calls, each immediately inside its own `std::lock_guard<std::mutex>` scope (L833, 857, 870, 889, 920, 931, 972, 998, 1013 of the committed file), plus the startup-failure `claimFailureTeardown` path; no cancellation helper can acquire `impl.mutex` while it is held |
| Windows removed by construction | the callback gate is installed in the same critical section as the run claim; the transport commit and the publication of `Running` are one critical section |
| New tests | `r3b1_stop_wins_at_the_earliest_moment_after_starting_is_published`, `r3b1_stop_wins_after_the_transport_startup_commits_before_running`, `r3b1_concurrent_start_and_stop_complete_without_deadlock` |
| Control-experiment residue | none — no `CONTROL VARIANT`, no debug print, no `sleep_for` in library code (the only `sleep_for` is the 200 µs poll interval of the test-side `RunningWatcher`) |
| Matches the previously validated baseline | clean build 0 warnings, `9/9` Core CTest targets, **71** Core cases (lifecycle 26) |

### 4–5. Commits pushed

| Commit | Content |
|---|---|
| `2975dc0` | `fix(core): remove the mutex re-entry from the start() cancellation paths` (session.cpp, +100/−69) |
| `bb9dfe9` | `test(core): cover the R3-B1 cancellation windows deterministically` (test_session_lifecycle.cpp, +212) |

Branch `core/21-session-frame-pipeline`, PR #22 (still Draft):

```
0a2ee8b..bb9dfe9  core/21-session-frame-pipeline -> core/21-session-frame-pipeline
PR #22 head = bb9dfe918e10c907efb6dd2a962839eeed9542ca
```

### 6. Verification against the committed head

Verified by exporting the commit into a clean tree (`git archive --format=zip -o head-bb9dfe9.zip bb9dfe9`
→ separate build directory), so nothing from the working tree could influence the result.

Environment: Windows 11 / x86_64, MSVC `14.44.35207` (x64), Ninja, `CMAKE_BUILD_TYPE=Release`,
Qt `6.8.3 msvc2022_64`, `HYREMOTE_BUILD_CORE=ON HYREMOTE_BUILD_TESTS=ON HYREMOTE_BUILD_SPIKES=ON
HYREMOTE_BUILD_EXAMPLES=ON`.

```bash
cmake -S build/verify-head -B build/verify-head-out -G Ninja \
      -DCMAKE_BUILD_TYPE=Release -DCMAKE_MAKE_PROGRAM=C:/Qt/Tools/Ninja/ninja.exe \
      -DCMAKE_PREFIX_PATH=C:/Qt/6.8.3/msvc2022_64 \
      -DHYREMOTE_BUILD_CORE=ON -DHYREMOTE_BUILD_TESTS=ON \
      -DHYREMOTE_BUILD_SPIKES=ON -DHYREMOTE_BUILD_EXAMPLES=ON
# Configure done in 6.6 s + 0.4 s generate. Only note: "Could NOT find WrapVulkanHeaders" (expected).

cmake --build build/verify-head-out
# 0 errors, 0 warnings in 8.9 s; second invocation: "ninja: no work to do".

ctest --test-dir build/verify-head-out
# 100% tests passed, 0 tests failed out of 23      Total Test time (real) = 41.64 sec
# (9 Core targets + 14 spike smoke tests, all Passed)
```

Core case counts at the committed head:

| Test executable | Cases |
|---|---|
| `hyremote-core-test-frame-lifetime` | 7 |
| `hyremote-core-test-damage` | 5 |
| `hyremote-core-test-timing` | 7 |
| `hyremote-core-test-mailbox` | 8 |
| `hyremote-core-test-transport-handoff` | 7 |
| `hyremote-core-test-session-lifecycle` | **26** (23 previously + the 3 R3 cases) |
| `hyremote-core-test-input-routing` | 5 |
| `hyremote-core-test-callback-lifetime` | 6 |
| **total deterministic cases** | **71** |

Stability: `ctest --test-dir build/verify-head-out -R hyremote-core` run 5× — `9/9` passed every time.

Content proof (run against the exported tree of `bb9dfe9`):

```
core/src/session.cpp: "static void markStartCancelledLocked"   -> L538   (present)
core/src/session.cpp: "static void finishCancelledStart"       -> absent
core/src/session.cpp: "markStartCancelledLocked(impl);"        -> 9 call sites, all inside lock scopes
core/tests/test_session_lifecycle.cpp: "HYR_TEST(r3b1..."      -> 3 cases present
"CONTROL VARIANT" / debug prints                              -> absent
```

### Verification history at a glance

| Head | Build | `ctest` | Core cases | Contains the R3 fix |
|---|---|---|---|---|
| `0a2ee8b` (previous head, verified in comment `5673459606`) | 0 warnings | 23/23 (39.43 s) | 68 | no |
| `bb9dfe9` (current head, this comment) | 0 warnings | 23/23 (41.64 s) | 71 | yes |

No code was changed for this verification; the local tree is clean apart from the two commits above.

skawu:
## Owner Review — Round 3: PASS / CORE-02 accepted

Reviewed current head `bb9dfe918e10c907efb6dd2a962839eeed9542ca` against the Round-3 blocker and the previously accepted Round-1/Round-2 semantics.

Accepted evidence:

- R3 fix commit `2975dc0` replaces the mutex-reentering cancellation helper with `markStartCancelledLocked()` whose contract requires the caller to hold `impl.mutex` and never re-locks it.
- All cancellation paths now either mark cancellation while already inside a lock scope or acquire the lock exactly once before marking it; the old `finishCancelledStart()` helper is absent.
- Transport commit and `Running` publication are one critical section, so a stop that wins before that point cannot be undone by a late publication of `Running`.
- Three R3 regression cases cover the earliest post-`Starting` cancellation, cancellation after transport startup before publication, and repeated concurrent start/stop completion without deadlock.
- Clean committed-head verification at `bb9dfe9`: Windows 11 x86_64 / MSVC 14.44 / Qt 6.8.3, 0 warnings, full root `ctest` 23/23, 71 deterministic Core cases; Core-only 9/9 repeated 5 times.
- Scope stayed inside #21; no Widgets/Quick adapter, transport backend, DMA-BUF/RKMPP or product-facade creep.

Round-1 B1-B4 and Round-2 teardown/start-cancellation corrections remain accepted and are not reopened.

**Disposition: PASS. PR #22 is ready to merge.**

After merge, #21 is complete and V0.0.1.0 execution advances to #44 (`HyRemote::RemoteAccess`) and #43 (SDK/install/export), then #6/#28/#29/#27/#11.

Issue #39, opened 2026-09-15
[V0.0.1.0][SDK] Freeze Qt-like SDK consumption, public facade and delivery contract


skawu:
## Product milestone

Supports **V0.0.1.0 — x86_64 (Windows + Linux) / Embedded C++ API** and defines the SDK/consumption contract reused by later QML/QPA milestones and V1.0.0.0 GA.

> **V1 convergence update (2026-09-16):** the implementation/package candidate is consolidated in Draft PR #106. This issue is the canonical SDK/user-consumption authority and is aligned with the final minimal V1 artifact model frozen by #101/#106.

Governance mode: `transitional-explicit` until #14 passes reusable ADS machine acceptance.

## Owner decision

HyRemote is an independent SDK/framework. It must not masquerade as a built-in Qt module, patch the user's Qt SDK, or require ordinary application developers to understand HyRemote internals.

Product goal:

> After HyRemote is installed or added as source once, normal use should feel like using a small Qt-oriented framework module.

Core, CaptureSource, RemoteFrame, Session, InputSink, Transport, bounded RFB, historical NeatVNC/rustvncserver work, QPA-private machinery and future DMA-BUF/hardware backends are implementation architecture, not normal application integration choices.

## Frozen installed V1 artifact model

The normal installed application surface is deliberately small:

- package: `HyRemote`;
- **one normal installed C++ imported target: `HyRemote::RemoteAccess`**;
- backing shared runtime: `HyRemoteRemoteAccess`;
- public facade header: `HyRemote/RemoteAccess.h`;
- optional QML module payload over the same runtime;
- optional exact-Qt `qhyremote` platform MODULE payload;
- one deployment helper: `hyremote_deploy(TARGET ... [QML] [QPA])`;
- CMake config/export/version metadata;
- required LICENSE/NOTICE/release metadata.

`hyremote-core` remains STATIC source/internal composition and is **not installed/exported as a normal V1 SDK target**. The installed package does not export `HyRemote::QpaPlatform` to applications. QML is not a second normal C++ product target.

`BUILD_SHARED_LIBS` must not turn these internal/product boundaries into an alternate application contract.

### 1. Prebuilt SDK — primary x86 developer path

Windows x86_64 and Linux x86_64 packages provide a standalone install prefix containing the applicable public/product artifacts above. The SDK must not require files to be written into the Qt installation tree.

[...]

PR #40, opened 2026-09-15
docs(sdk): freeze user consumption and delivery contract (#39)

skawu:
## Summary

Implements the accepted product correction from #39 at repository-documentation level.

- defines HyRemote as an independent SDK that should feel like a Qt module after acquisition;
- freezes two first-class consumption paths: prebuilt SDK and source/CMake integration;
- defines the normal product target as `HyRemote::RemoteAccess`;
- defines the target C++ facade shape for Widgets and Quick;
- records the later `import HyRemote` QML direction;
- makes deployment support a product responsibility instead of a manual backend-copy task;
- explicitly keeps transport/capture/runtime/toolchain choices behind the public product boundary;
- makes examples and user guides required release-acceptance artifacts;
- removes README wording that prematurely presents one VNC backend as the product-facing choice.

## Important scope

This PR does **not** claim that `HyRemote::RemoteAccess`, the install/export package, deploy helper or examples are implemented yet. #39 remains open until those are real and validated on Windows/Linux.

The purpose of this PR is to make the user-consumption contract canonical before implementation continues, so Core/adapter/transport work converges toward one stable product surface.


skawu (commented):
Owner self-review PASS. Scope is documentation-only and correctly freezes the accepted SDK/user-consumption contract without claiming unimplemented runtime functionality. `HyRemote::RemoteAccess`, standalone SDK/source consumption, deployment ownership, backend invisibility, examples and user documentation are now the product boundary that later code must satisfy.

PR #46, opened 2026-09-15
build(sdk): add install/export package and external consumers (#43)

skawu:
## Product context

Implements the package/distribution foundation of #43 under **V0.0.1.0 — Windows + Linux x86_64 / Embedded C++ API**.

PR #45 has merged. This PR is now retargeted directly to `main`; review scope is only the SDK/install/export increment.

## Scope

- export `HyRemote::Core` and `HyRemote::RemoteAccess` from one installable CMake package;
- `HyRemoteConfig.cmake` + version config + `HyRemoteTargets.cmake`;
- install public headers, generated export header, libraries and Apache-2.0 license;
- installed public target names remain `HyRemote::Core` / `HyRemote::RemoteAccess`;
- initial installed `hyremote_deploy(TARGET ...)` hook delegates Qt runtime deployment to Qt's supported CMake deployment API and reserves the stable hook for future HyRemote transport/plugin payloads;
- clean external installed-SDK consumer fixture;
- clean `add_subdirectory()` source-consumer fixture using the same `HyRemote::RemoteAccess` target;
- Windows x86_64 + Linux x86_64 CI that builds/installs the SDK, builds/runs the installed consumer, then builds/runs the source consumer.

## Evidence

The stacked branch has already passed installed/source consumer validation on both Windows x86_64/MSVC and Linux x86_64/GCC with Qt 6.8.3. After retargeting to `main`, the PR remains Draft until the direct-main diff and fresh CI are rechecked.

## Boundaries

This does not select a VNC backend, add target adapters, or claim V0.0.1.0 complete. Runtime transport/plugin deployment is extended under #27 when a production backend exists. #43 remains open until the complete SDK payload and product paths are accepted.

Governance mode: `transitional-explicit`.

Issue #29, opened 2026-09-15
[V0.0.1.0][INPUT] Implement normalized Qt remote input for x86 Windows + Linux


skawu:
## Product milestone

Supports **V0.0.1.0 — x86_64 (Windows + Linux) / Embedded C++ API** and the shared input model reused by later QML/QPA modes.

> **V1 convergence update (2026-09-16):** repository implementation has been absorbed into Draft PR #106, the only active V1 product line. This issue is now the normalized-input correctness/acceptance authority; it does not authorize a separate implementation branch or a new runtime policy API.
>
> **Terminal input-state correction (2026-09-16):** #106 now also treats explicit HyRemote runtime stop as an input-lifecycle boundary. After transport/Core callbacks are quiescent, the target input adapter discards remote input that never reached the Qt GUI and balances remote key/button state that was already delivered. This is an internal target-adapter/runtime guarantee; it adds no application API and is shared by C++, QML and QPA through the same `RemoteAccess` runtime.

Governance mode: `transitional-explicit` until #14 passes reusable ADS machine acceptance.

## Goal

Provide the minimum transport-neutral remote-input model required by the product and Qt delivery adapters that behave consistently on **Windows x86_64 and Linux x86_64**.

Normal Embedded C++ integration injects normalized events directly into the target Qt application. OS-level virtual input (`uinput`, Windows virtual HID, desktop-wide injection, etc.) is not the V1 default path.

## Required normalized events

At minimum preserve transport-neutral representation for:

- pointer move / absolute position;
- pointer button press/release;
- wheel/scroll;
- key press/release;
- modifier state;
- text/unicode input where the protocol/viewer provides enough information;
- target/frame geometry metadata required for coordinate mapping.

Do not expose RFB keysyms, Windows VK codes, Linux evdev codes, socket/client objects or backend-specific structs as the public Core/application contract.

## Qt delivery requirements

Provide target adapters for the explicit product path:

- `QWidget` / Widgets;
- `QQuickWindow` / Qt Quick.

Delivery must be marshalled to the appropriate Qt GUI thread without synchronously blocking transport/network work. QML reuses the same Quick/product path; it does not create another input stack.

[...]

Issue #27, opened 2026-09-15
[V0.0.1.0][TRANSPORT] Select and implement x86 Windows + Linux VNC/RFB transport baseline


skawu:
## Product milestone

Supports **V0.0.1.0 — x86_64 (Windows + Linux) / Embedded C++ API**.

Governance mode: `transitional-explicit` until #14 passes final ADS machine acceptance.

## Product/user-consumption authority

**#39 is now the canonical SDK and public-consumption contract.**

Transport is an internal implementation choice. A backend is not acceptable merely because it compiles or passes protocol tests; it must fit the product consumption model:

- normal applications link/use `HyRemote::RemoteAccess`, not backend targets;
- no backend type appears in the normal public C++ or QML API;
- prebuilt Windows/Linux SDK consumers must not install/configure a backend-specific toolchain such as Rust/Cargo;
- source-build and future cross-compilation burden is a first-class selection criterion;
- changing the default transport must not require application-source changes;
- runtime deployment of the selected backend is owned by HyRemote packaging/deployment support rather than by the application developer.

This criterion supersedes any tendency to optimize backend selection before user-facing integration is frozen.

## Why this issue exists

The product baseline requires **both Windows x86_64 and Linux x86_64** before V0.0.1.0 can complete.

The existing #4 conclusion is a valid **Conditional GO for NeatVNC as a Linux-oriented first RFB transport candidate**, but it does not prove Windows support. Therefore HyRemote must not simply implement NeatVNC and call the x86 milestone complete.

## Goal

Freeze and implement a production VNC/RFB transport strategy that:

- works on Windows x86_64 and Linux x86_64;
- preserves one transport-neutral `HyRemote::Core` contract;
- is fully hidden behind the #39 product facade/package contract;
- does not leak backend types into public Core/API headers;
- is compatible with HyRemote's Apache-2.0 distribution model;
- supports standard VNC viewers with at least a universal baseline encoding;
- delivers normalized pointer/keyboard events back into HyRemote;
- owns bounded downstream queue/backpressure so slow clients never block Core capture/render paths;
- does not impose disproportionate build, packaging or embedded cross-compilation complexity on downstream users.

### NeatVNC

Status: **keep as Linux-oriented candidate / accepted #4 input**.

Strengths already recorded in `docs/neatvnc-evaluation.md`: ISC license, frame/damage APIs, input callbacks, optional GBM/H.264, thin adapter fit.

[...]

Issue #6, opened 2026-09-14
[V0.0.1.0][WIDGETS] Deliver x86 Windows + Linux Qt Widgets Embedded C++ API path


skawu:
## Product milestone

Supports **V0.0.1.0 — x86_64 (Windows + Linux) / Embedded C++ API**.

Governance mode: `transitional-explicit` until #14 passes final ADS machine acceptance.

## Goal

Deliver the first end-to-end HyRemote product path for a normal Qt Widgets application through the shared Embedded C++ API on **both Windows x86_64 and Linux x86_64**.

Passing on only one operating system is partial evidence, not completion.

## Required flow

```text
QWidget target
  -> Widgets target adapter
  -> portable x86 capture backend
  -> RemoteFrame
  -> supported VNC/RFB transport backend
  -> standard VNC viewer

Viewer input
  -> transport callbacks
  -> normalized HyRemote input
  -> Qt Widgets input adapter
  -> QWidget application
```

The public application-facing API/session semantics must be the same on Windows and Linux even if the selected transport or low-level capture implementation differs by OS.

## Scope

- minimal `examples/widgets-basic` product example;
- Embedded C++ integration API;
- Windows x86_64 support;
- Linux x86_64 support;
- full-frame correctness baseline first;
- mouse/pointer input;
- keyboard input;
- local UI remains usable while remote viewing/control is active;
- disconnect/reconnect without application restart;
- bounded frame/backpressure path through `HyRemote::Core`;
- initial runtime diagnostics/logging;
- compatibility entry for each OS.

## Validation

Validate independently on Windows x86_64 and Linux x86_64:

- child widgets;
- menus/popups/dialogs where supported by the chosen target/capture path;
- QPainter custom drawing;
- resize/geometry handling;
- remote input coordinate mapping;
- keyboard focus/input;
- local + remote input coexistence;
- repeated connect/disconnect;
- memory stability;
- a slow remote client does not stall the Qt GUI indefinitely.

## Product API requirement

The example must demonstrate the intended explicit C++ developer experience rather than wiring internal Core interfaces directly from application code. Internal adapters may implement `CaptureSource`, `Transport` and `InputSink`, but those implementation details must not become the normal application integration contract.

## Non-goals

- Declarative QML API (`V0.0.2.0`);
- Transparent QPA Proxy (`V0.0.3.0`);
- QOpenGLWidget performance optimization;
- embedded EGLFS claims;
- hardware H.264;
- DMA-BUF/zero-copy.

[...]

PR #47, opened 2026-09-15
feat(input): define the V0.0.1 normalized input contract (#29)

skawu:
## Product context

First implementation slice of #29 for **V0.0.1.0 — Windows + Linux x86_64 / Embedded C++ API**.

This PR freezes the protocol-neutral Core vocabulary and geometry semantics before protocol/Qt adapters consume it. #29 remains open after this slice; actual QWidget/QQuickWindow event delivery is completed with #6/#28 and VNC normalization with #27.

## Scope

- replace the placeholder Core input schema with transport-neutral pointer/key/text/modifier events;
- define the minimum V0.0.1 logical key vocabulary without Qt/VNC/OS numeric codes;
- define source viewport + DPR metadata and edge-preserving pointer mapping to target logical coordinates;
- make invalid/non-finite geometry fail explicitly;
- keep committed UTF-8 text separate from logical key identity;
- preserve Core routing/threading semantics: transport callback -> InputSink::post(), no synchronous GUI wait;
- deterministic unit tests for geometry, high-DPI mapping, clamping, modifiers, key/text/scroll separation;
- document the adapter boundary and deferred responsibilities.

## Integration status

The branch has been reconstructed directly on current `main` after PR #46. The SDK/install/export rules and Windows shared-DLL export fix are preserved; the only `core/CMakeLists.txt` product delta here is adding `src/input.cpp` to the existing exported Core target.

## Non-goals

- no VNC keysym normalization (#27);
- no QWidget/QQuickWindow Qt event injection yet (#6/#28);
- no uinput/virtual-HID;
- no IME composition parity;
- no change to view/input authorization policy.


skawu:
Owner Review — PASS

Accepted normalized-input contract on head `bbd0164a9aaefd6caba351dd4f142ad1f2354ffe`.

Verified:
- branch is rebuilt directly on current `main` after #46; SDK/install/shared-DLL rules are preserved;
- Core exposes protocol/Qt/OS-neutral pointer, button, scroll, key, modifier and committed-text vocabulary;
- modifier and lock keys have explicit `KeyCode` identities as well as state-mask semantics, so press/release transitions do not need to be guessed by target adapters;
- pointer mapping is explicit, edge-preserving, clamps out-of-range input, rejects invalid/non-finite geometry, and does not double-apply DPR;
- committed UTF-8 text remains separate from logical keys and is documented as sensitive/not logged by default;
- Core routing/threading/lifetime semantics remain unchanged;
- Windows x86_64/MSVC + Qt 6.8.3 regression PASS;
- Linux x86_64/GCC + Qt 6.8.3 regression PASS;
- installed/source SDK consumption PASS on both OSes.

Scope remains bounded: this merge does not claim QWidget/QQuickWindow input injection or VNC keysym normalization. #29 remains open for those downstream product slices.

PR #48, opened 2026-09-15
feat(widgets): add built-in QWidget capture and input adapter (#6)

skawu:
## Product context

Production Qt Widgets target integration for #6 under **V0.0.1.0 — Windows + Linux x86_64 / Embedded C++ API**, on the path to V1.0.0.0.

This PR is reconstructed directly on current `main` after #46 and #47. It preserves the install/export/shared-DLL SDK baseline and the normalized Core input contract.

## Scope

- built-in Qt Widgets target adapter selected internally from `QObject*`;
- asynchronous GUI-thread capture using public `QWidget::render()`;
- each published frame owns `CpuFrameStorage` pixels for the full `RemoteFrame` lifetime;
- RGBA8888 premultiplied CPU baseline, DPR-aware pixel dimensions, Completion PTS, FullFrame damage;
- queued capture requests become no-ops after stop and callbacks are drained before `stop()` returns;
- destroyed QWidget target reports non-recoverable `TargetLost`;
- normalized QWidget input sink for pointer move/button/wheel, key press/release, modifiers and committed UTF-8 text;
- input is always queued to the Qt GUI thread and never synchronously blocks the transport callback;
- queued input is dropped when the sink is destroyed;
- installed `HyRemoteConfig.cmake` declares Qt Widgets when the built package contains this adapter;
- deterministic headless Widgets tests;
- Windows x86_64 + Linux x86_64 Qt 6.8.3 CI plus the existing SDK consumption regression.

## Boundaries

This is the correctness/product baseline, not the final performance ceiling. It does not add production VNC transport, multi-top-level composition, damage optimization, external-buffer acceleration, or full IME composition parity. Those remain separate WBS toward V1.0.0.0.


skawu:
Owner Review — PASS

Widgets product baseline accepted on head `a92c7a12a11c81e47721b15f7741ea5cb8a0693d`.

Verified:
- branch reconstructed directly on current main after #46/#47;
- QWidget capture remains GUI-thread-only and publishes owned RGBA8888 premultiplied `RemoteFrame` storage with Completion PTS;
- normalized QWidget input sink delivers pointer/button/wheel, key press/release, modifiers and committed UTF-8 text through queued GUI-thread work;
- queued input is dropped after sink destruction; queued capture is suppressed after stop; target destruction is surfaced as `TargetLost`;
- installed package declares Qt Widgets when this adapter is present;
- Windows x86_64/MSVC + Qt 6.8.3: Widgets, RemoteAccess and SDK consumption PASS;
- Linux x86_64/GCC + Qt 6.8.3: Widgets, RemoteAccess and SDK consumption PASS.

This closes the production QWidget target-adapter slice. Remaining #6 compatibility/performance/examples can continue as V1.0.0.0 WBS without blocking Quick/VNC work.

PR #54, opened 2026-09-15
refactor(core): share the capability list formatter and pin its reason strings

skawu:
## Scope

Clean up `core/src/capabilities.cpp` and lock down the behaviour it produces. Two commits:

- `c7bff09` `refactor(core): share the capability list formatter and include the headers used`
- `d13b3a6` `test(core): pin the compatibility reason strings verbatim`

## What changed

| File | Change |
| --- | --- |
| `core/src/capabilities.cpp` | `formatList()` and `domainList()` were two structurally identical copies of the same loop, and both appended to the result string without reserving. They now delegate to one `joinNames()` template that computes the exact result length (separators plus name lengths) and reserves it before building, so the string never reallocates. The thin adapters keep the call sites and the user-visible strings unchanged. The translation unit also includes the standard headers it uses directly (`<cstddef>`, `<string>`, `<string_view>`, `<vector>`) instead of relying on `capabilities.hpp` to provide them transitively. |
| `core/tests/test_transport_handoff.cpp` | The compatibility reasons were only covered by substring assertions, so the joined-list format - exactly what the refactor touches - was untested. The suite now asserts the verbatim reason for a two-format-per-side CPU mismatch and for a single-versus-multiple external storage domain mismatch. |

## Impact

- **API / ABI**: none. No public header, signature, enum or default value changed.
- **Behaviour**: none. The exact reason strings are now asserted, and the assertions pass against both the previous implementation (temporarily restored from `main`) and the new one, so the refactor is proven output-preserving.
- **Boundaries / dependencies**: none. No new dependency, no Qt or protocol involvement in `hyremote-core`.
- **Performance**: unchanged in practice (the lists hold a handful of entries); the formatter simply no longer grows the string incrementally.

## Verification

| Check | Result |
| --- | --- |
| Build | MSVC 14.44 x64, Ninja, `Release`: 0 errors, 0 warnings |
| `ctest --test-dir build/core` | 12/12 passed (80 deterministic Core cases) |
| `ctest --test-dir build/all` (Core + tests + spikes) | 26/26 passed, 41.9 s |
| Output equivalence | the new assertions pass with the old and with the new implementation |
| clangd | `core/src/capabilities.cpp` checks clean against the project compilation database |

Tested on Windows 11 x86_64 with Qt 6.8.3 (`msvc2022_64`); Linux x86_64 is covered by CI.

## Non-scope

No change to the capability contract itself, to `Session` / `RemoteFrame` / dispatch, to capture or transport backends, or to the QML/QPA work. No Issue is linked: this is a review follow-up on already-merged Core code rather than a planned capability.

## Known limitations

None identified. The two adapters remain only to keep the call sites readable.


skawu (commented):
Owner review: PASS. The change is bounded to Core diagnostic formatting hygiene plus regression coverage, introduces no Qt/protocol/backend dependency, changes no public API/ABI, and preserves the existing compatibility-reason text. Exact-string assertions are acceptable here because they lock behavior touched by the refactor without changing the architecture contract. Current head had successful Windows/Linux product CI on its submitted baseline. Ready to squash-merge.

skawu:
QML line continues under the frozen V0.0.2 product contract. No local Agent is required for repository implementation or documentation work; hosted-runner non-execution remains infrastructure evidence only and does not stop parallel V1.0 work.

PR #55, opened 2026-09-15
docs: make the root README an entry point and record the product overview

skawu:
## Scope

Repository front page and developer experience, now synchronized onto current `main`.

Final diff is intentionally limited to:

- `README.md` — conventional project entry point, quick build/use paths, product status, docs index and support boundaries;
- `docs/product-overview.md` — long-form product narrative moved out of the root README;
- `spikes/capture/README.md` — replace the reference-machine Qt path with a portable placeholder.

The earlier `.clangd` experiment is no longer part of this PR.

## Product/consumption corrections incorporated during review

- Qt wording now keeps **Qt 6.8.x** as the current V1 reference/support line instead of creating an unverified standing `6.8+` promise;
- installed SDK and source/vendored consumption are explicitly separated:
  - installed SDK → `find_package(HyRemote CONFIG REQUIRED)`;
  - source tree → `add_subdirectory(...)`, then link the same `HyRemote::RemoteAccess` target;
- the README no longer implies that `find_package(HyRemote)` follows automatically from source inclusion;
- safe defaults, Windows/Linux peer status, evidence-based compatibility, three mandatory V1 integration modes, Apache-2.0/trademark boundary and post-V1 embedded direction remain unchanged.

## Why

The previous root README was primarily a long product brief. The new root is an evaluator/developer entry point while the full product framing lives in `docs/product-overview.md`.

## Review/verification boundary

This is documentation-only. It does not change CMake targets, runtime behavior, public API, transport, security semantics or product milestones. The branch is now based on current `main`; its diff contains only the three files listed above.


skawu (commented):
Owner review — CHANGES REQUIRED before merge.

The overall README/product-overview restructuring is useful and aligned with the V1 product framing, and the previous `.clangd` concern has been removed from the current diff. One remaining product-consumption defect is still present and blocks merge:

**B1 — installed-SDK and source-consumption instructions are still conflated.** The README says “After installing HyRemote (or adding it as a source subdirectory), consumption is the same as for a Qt module” and immediately shows `find_package(HyRemote CONFIG REQUIRED)`. A vendored/add_subdirectory consumer does not obtain an installed `HyRemoteConfig.cmake`; the repository's canonical source-consumer path links the target after `add_subdirectory(...)` instead. Split this section into two explicit paths: (a) installed SDK → `find_package(HyRemote CONFIG REQUIRED)`; (b) source/vendored → `add_subdirectory(...)` then link `HyRemote::RemoteAccess`. Do not imply `find_package(HyRemote)` after source inclusion.

Also sync onto current `main` after the correction so the front-page status reflects the current merged RFB/security baseline and does not overwrite newer product wording. Keep scope to README/product-overview/spike-path cleanup; no architecture or milestone changes.

No local Agent is required for this correction.

skawu (commented):
Owner review — PASS.

The blocking consumption defect is corrected: installed-SDK package discovery and source/vendored `add_subdirectory()` are now explicit separate flows that converge on the same `HyRemote::RemoteAccess` target. The branch is synchronized to current `main`, the earlier `.clangd` experiment is absent, and the final diff is limited to README/product overview/spike-path documentation.

Product boundaries remain frozen: Qt 6.8.x is the V1 reference line, the three V1 integration modes are unchanged, support remains evidence-based, and no runtime/API/security behavior is modified. Ready for squash merge.

Issue #53, opened 2026-09-15
[V0.0.1.0][TRANSPORT] Implement bounded cross-platform RFB 3.8 baseline behind RemoteAccess


skawu:
## Parent

#27 — production VNC/RFB transport baseline.

## Why a HyRemote-owned baseline is now authorized

The frozen rule was to avoid a custom RFB implementation unless a bounded evaluation showed no acceptable maintained permissive backend for the required Windows + Linux x86 product path.

That condition is now met for the V0.0.1/V1.0 critical path:

- NeatVNC remains the accepted Linux/Embedded-Linux-oriented candidate, not a Windows backend;
- rustvncserver protocol/FFI viability passed #38, but #52 closed **NO-GO for the common production backend** because released explicit-bind/start-ready semantics are insufficient and upstream input/event paths are unbounded;
- libvncserver and other established server stacks are GPL-family and incompatible with the default Apache-2.0 product distribution goal;
- `oxidecomputer/rfb` is MPL-2.0 and its current server trait exposes key input but discards pointer events, so meeting V0.0.1 input requirements would require maintaining an upstream modification while also adding file-level copyleft compliance;
- other surveyed candidates are stale, runtime/toolchain-mismatched, provenance-poor, or introduce a second product runtime/build system.

The product must not wait indefinitely for an upstream backend change. A deliberately small internal RFB correctness baseline is therefore allowed, while richer Linux/embedded transports remain replaceable backends.

## Architecture boundary

Implement an **internal transport adapter**, not a new public protocol API.

```text
HyRemote::RemoteAccess
  -> createDefaultTransport()
  -> bounded RFB transport adapter
  -> standard VNC viewer
```

Hard invariants:

- no RFB/VNC type in `hyremote-core` or public `HyRemote/RemoteAccess.h`;
- no application backend selection in the normal path;
- Qt Widgets and Qt Quick share the same transport;
- transport may use Qt Network because the product facade already depends on it; Core stays Qt-free;
- construction remains inert; bind occurs only during explicit `start()`;
- loopback remains the default bind address;
- `enqueueFrame()` is bounded/nonblocking and keeps only freshness-oriented latest-frame state;
- no unbounded per-client frame/request queue;
- slow clients never block Core/Qt render threads;
- `stop()` is quiescent: callbacks cannot fire after return.

## V0.0.1 protocol scope

Minimum interoperable RFB 3.8 server sufficient for the product acceptance path:

- protocol 3.8 handshake;
[...]

Issue #52, opened 2026-09-15
[V0.0.1.0][TRANSPORT-UPSTREAM] Resolve rustvncserver production blockers or reject it as the common x86 backend


skawu:
## Parent

#27 — production VNC/RFB transport baseline for Windows + Linux x86_64.

#38 / PR #50 is complete with **CONDITIONAL GO** evidence only. This child converts the remaining upstream-dependent blockers into one bounded decision task; it does not reopen the completed protocol/product-fit spike.

## Frozen architecture constraints

- public product API remains `HyRemote::RemoteAccess` / later `import HyRemote`;
- no Rust/Tokio/backend type in Core or public product headers;
- construction never opens a listener; normal default is loopback;
- `Transport::enqueueFrame()` and all HyRemote-owned downstream queues remain bounded/nonblocking toward Core;
- slow or malicious remote clients must not stall Qt GUI/rendering indefinitely;
- prebuilt SDK consumers must not install/configure Rust/Cargo;
- no permanent HyRemote-only fork of an upstream protocol backend;
- custom RFB remains last resort only.

## Accepted evidence from #38

On Windows and Linux x86_64, the candidate can:

- build behind an opaque HyRemote C ABI;
- interoperate with maintained `vncdotool`;
- deliver known framebuffer pixels;
- deliver pointer/key events;
- disconnect/reconnect;
- stop and release its listener deterministically.

The HyRemote-owned probe queue is bounded. This does **not** make upstream internals bounded.

### U1 — released explicit bind / ready contract

Current released v2.2.1 binds wildcard internally. Upstream PR #29 demonstrates `listen_on(SocketAddr)` but is not merged/released, and bind readiness is still represented by a long-running future rather than an explicit pre-bound/ready object/result.

Production acceptance requires an upstream/released or equivalently stable supported API for:

- exact bind address;
- deterministic occupied-port/bind failure;
- no false success because an accept task was merely scheduled;
- deterministic stop ownership.

### U2 — bounded event channels

Current upstream main uses `mpsc::unbounded_channel()` for both server events and per-client events. The exact source shape is now verified under `src/server.rs` / `src/client.rs`.

Production acceptance requires an upstream-supported bounded strategy that preserves input correctness. It is not sufficient to put a bounded queue only after the upstream unbounded queues.

Preferred semantic direction:

- bounded per-client message/event path;
- backpressure confined to that client's network/runtime task, never Qt/Core;
[...]

Issue #30, opened 2026-09-15
[V0.0.1.0][ACCEPTANCE] Freeze dual-OS x86 Embedded C++ API release gate


skawu:
## Product milestone

This is the final product acceptance gate for **V0.0.1.0 — x86_64 (Windows + Linux) / Embedded C++ API**.

Governance mode: `transitional-explicit` until #14 passes reusable ADS machine acceptance.

> **V1 convergence authority update (2026-09-16):** implementation is absorbed into the single V1 line #106. The C++ product/package contract is now frozen by #39/#101/#106: one normal installed C++ target `HyRemote::RemoteAccess`, one shared `HyRemoteRemoteAccess` runtime, Core statically composed/internal, and one `hyremote_deploy()` entry point.

## Purpose

Prevent V0.0.1.0 from being declared complete because individual technical modules compile, because only one operating system works, or because internal Core/transport pieces exist without a usable external-developer integration path.

This issue closes only when the user-facing Embedded C++ capability is reproducibly consumable on both Windows x86_64 and Linux x86_64.

## Canonical user-consumption contract

**#39 is mandatory for V0.0.1.0.**

HyRemote is an independent SDK/framework whose normal usage should feel like a Qt module after it has been installed or added as source once.

Normal applications must not manually assemble internal `CaptureSource`, `Transport`, `Session`, mailbox, backend runtime, protocol-specific objects or Core targets.

The frozen product-level CMake/API shape is:

```cmake
find_package(Qt6 REQUIRED COMPONENTS Widgets)
find_package(HyRemote CONFIG REQUIRED)

target_link_libraries(MyApp PRIVATE
    Qt6::Widgets
    HyRemote::RemoteAccess
)
```

```cpp
#include <HyRemote/RemoteAccess.h>

HyRemote::RemoteAccess remote(&window);
remote.start();
```

Normal installed consumers do not link `HyRemote::Core`; Core is static/internal composition behind the shared product facade.

## Frozen Embedded C++ behavior

The same `HyRemote::RemoteAccess` facade supports qualified Qt Widgets (`QWidget`) and Qt Quick (`QQuickWindow`) targets through internal adapters.

Product invariants:

- construction is inert and opens no listener;
- explicit `start()` / `stop()` lifecycle;
- default listen address is loopback (`127.0.0.1`);
- default port is 5900;
- remote input is disabled by default;
[...]

PR #56, opened 2026-09-15
feat(transport): add bounded cross-platform RFB baseline (#53)

skawu:
## Product context

Implements the production correctness transport authorized by #53 under #27 for **V0.0.1.0 — x86 Windows + Linux / Embedded C++ API**, on the continuous path to V1.0.0.0.

The custom baseline is now permitted only because the bounded backend survey is complete: rustvncserver is NO-GO for the current common production path (#52), NeatVNC is Linux-oriented, and the remaining surveyed stacks fail the frozen license/input/maintenance/toolchain constraints.

## Architecture

- RFB remains entirely internal to `HyRemote::RemoteAccess`; no protocol/backend type enters Core or the public facade.
- same transport serves Widgets and Quick;
- Qt Network is reused instead of introducing another runtime/toolchain;
- construction remains inert; exact bind occurs only in `Transport::start()`;
- Core handoff is latest-frame-wins with exactly one retained pending `RemoteFrame`;
- update requests are coalesced per client and no new framebuffer is queued while that client's socket has pending output;
- client count and input protocol buffer are bounded;
- `stop()` synchronously tears down the transport thread/listener/clients before returning.

## RFB baseline

- RFB 3.7/3.8 handshake;
- SecurityType None correctness baseline with loopback-safe product default (Internet exposure remains unsupported);
- ServerInit;
- bounded SetPixelFormat / SetEncodings parsing;
- Raw framebuffer rectangles from the existing `Rgba8888` x86 capture contract;
- DesktopSize when the viewer advertises it; otherwise resize fails explicitly and requires reconnect;
- normalized pointer/button/wheel/key/text input;
- disconnect/reconnect;
- max 8 simultaneous clients.

Compression, TLS, clipboard and hardware encoding are intentionally outside this correctness slice.

## Evidence

Adds a dual-OS Qt 6.8.3 workflow using maintained `vncdotool` to verify:

- occupied-port bind failure;
- known framebuffer pixels;
- pointer/button/wheel/key/text input;
- disconnect/reconnect;
- explicit `Transport::stop()` and listener release.

The PR stays Draft until both Windows and Linux product-fit jobs are green and the boundedness/security review is complete.


skawu (commented):
Owner review — **PASS for the V0.0.1 transport slice**, with one responsibility boundary frozen for the next WBS.

Checked against #27/#39/#53 invariants and the dual-OS CI evidence:

- RFB implementation is internal to `remoteaccess`; no protocol/backend type enters `hyremote-core` or `HyRemote/RemoteAccess.h`.
- Construction is inert; exact bind happens synchronously during `Transport::start()`, so occupied-port failure is real and deterministic.
- Core handoff is latest-frame-wins with one retained frame and a coalesced worker notification; network encoding/writes remain on the transport thread.
- Per-client framebuffer requests are coalesced, a new update is not appended while `bytesToWrite()!=0`, protocol input accumulation is capped, and active clients are capped at 8.
- `stop()` uses blocking worker shutdown + thread join and clears callbacks before return, satisfying transport quiescence.
- Raw RFB baseline, pixel conversion, resize/DesktopSize handling, reconnect and normalized pointer/key/text callbacks are sufficient for the correctness baseline.
- Windows + Linux Qt 6.8.3 jobs are green with maintained `vncdotool`; the public SDK/facade and Widgets/Quick workflows are also green at this head.

**Do not overclaim input-delivery backpressure from this PR.** The transport protocol buffer/client/frame queues are bounded here, but the Qt target adapters still marshal each normalized input event into the GUI event loop. Flood/coalescing/rate policy for that final transport→InputSink→GUI leg remains part of #29 and must be accepted before #30/V0.0.1.0 closes. This is a WBS ownership clarification, not a relaxation of the product requirement.

No local-Agent evidence is required for this PR because GitHub-hosted Windows/Linux jobs already cover the claimed x86 build/interoperability path. Ready to merge after this review.

PR #59, opened 2026-09-15
fix(security): bound remote input delivery and expire incomplete handshakes (#58)

### LINKED ISSUE #58: [V0.0.1.0][SECURITY] Bound GUI input delivery and expire incomplete RFB handshakes
## Context

Post-merge automated review of PR #56 found two valid P1 issues in the V0.0.1.0 RFB correctness baseline. They were reported after #56 had already been squash-merged, so this issue owns the mandatory hot-correction on the continuous V1.0.0.0 path.

## P1-A — bounded transport→InputSink→GUI delivery

Current Widgets/Quick sinks queue one Qt invocation per normalized remote input event. A reachable client can flood pointer events faster than the GUI thread consumes them, causing unbounded Qt event-queue growth even though the RFB socket/protocol/frame queues are bounded.

Required correction:
- fixed-capacity input mailbox per target sink;
- freshness-oriented pointer-move coalescing;
- at most one pending GUI drain invocation per sink;
- key/button/text lifecycle events are not silently overwritten merely to preserve pointer motion;
- full lifecycle-sensitive mailbox fails in a bounded/recoverable way rather than growing memory;
- queued input is discarded after sink/session teardown.

## P1-B — incomplete handshake slot exhaustion

The RFB server caps clients at 8, but an incomplete handshake could occupy a slot forever. Eight idle sockets could therefore deny all legitimate viewers indefinitely.

Required correction:
- bounded handshake deadline for every accepted pre-Normal client;
- expired handshake is aborted and the slot released;
- normal connected clients are not expired by this timer;
- deterministic evidence that exhausted incomplete slots recover and a legitimate viewer can subsequently connect.

## Required evidence

Windows x86_64 + Linux x86_64 / Qt 6.8.3:
- Widgets input flood/coalescing regression;
- Quick input flood/coalescing regression;
- handshake-slot exhaustion/expiry regression;
- existing RFB vncdotool frame/input/reconnect/stop product-fit remains green;
- existing RemoteAccess/SDK/Widgets/Quick regressions remain green.

## Architecture constraints

No public API change. No RFB type in Core or `HyRemote/RemoteAccess.h`. Keep local UI responsive and preserve the frozen bounded/nonblocking design. This issue blocks #30 V0.0.1.0 acceptance but does not pause parallel QML/QPA development.

skawu:
## Context

Hot-correction for two valid P1 findings reported after PR #56 had already been squash-merged. This PR is intentionally based on the merged transport baseline and contains only the mandatory boundedness/availability corrections plus regressions.

## Corrections

- Widgets and Quick `InputSink` delivery now use a fixed-capacity mailbox instead of one Qt queued invocation per remote packet.
- Adjacent pointer motion is latest-value coalesced; when full, stale pointer motion is sacrificed before lifecycle-sensitive input.
- At most one GUI drain invocation is pending per sink; lifecycle-only overflow fails bounded/recoverably instead of growing memory.
- Sink teardown clears queued input.
- Every accepted RFB client has a 3 s pre-Normal handshake deadline; silent/incomplete peers are aborted so the max-eight-client limit cannot be held indefinitely.

## Regression evidence added

- 10,000-event pointer burst without GUI pumping must result in one freshest Widgets motion delivery.
- same test for QQuickWindow.
- eight silent RFB peers must all expire, after which a maintained `vncdotool` viewer must connect and capture the framebuffer.
- existing dual-OS RFB product-fit, Widgets, Quick, RemoteAccess and SDK workflows remain required.

## Frozen boundaries

No public API change. No RFB/backend type enters Core or `HyRemote/RemoteAccess.h`. Construction remains inert, loopback remains the product default, and Core/GUI threads remain isolated from network backpressure.

Closes #58 after dual-OS CI and review pass.


skawu (commented):
Owner review — **PASS** for #58 hot-correction.

Verified against the two post-merge P1 findings from #56 and the frozen bounded/nonblocking design:

- Widgets and Quick no longer enqueue one Qt invocation per packet; each sink owns a fixed 64-event mailbox and at most one pending GUI drain.
- pointer motion is freshness-coalesced / sacrificed before lifecycle-sensitive input; a lifecycle-only full mailbox fails at the existing transport callback boundary instead of growing memory;
- sink destruction clears queued input and queued drains become inert;
- incomplete RFB clients have a fixed 3 s pre-Normal deadline, while Normal clients are exempt;
- the eight-silent-client exhaustion regression proves slots recover and a standard viewer can subsequently connect;
- Windows + Linux RFB, Widgets, Quick, RemoteAccess and SDK workflows are all green at exact head `2d6161b0eb6e423e6ab9bde81cabc353a9c1948d`.

No public API/Core protocol leakage and no local-Agent evidence is required. Ready to squash merge.

Issue #60, opened 2026-09-15
[SUPERSEDED by #238][HISTORICAL][E2E-01] Public-facade Widgets/Quick examples


skawu:
## Parent / goal

Closes the largest remaining product-evidence gap across #6, #28, #29, #39 and #30 while feeding the richer V1.0 examples/docs program in #41.

Deliver real **Embedded C++ API** examples that use only `HyRemote::RemoteAccess`, then validate them end-to-end with a maintained standard VNC client on Windows x86_64 and Linux x86_64.

## Frozen design constraints

- no application assembly of Core/Session/CaptureSource/InputSink/Transport;
- one public facade for Widgets and Quick;
- construction remains inert, loopback-safe, remote input off by default;
- enabling control for the acceptance harness is explicit;
- local Qt window/render path remains the native application path; HyRemote is additive;
- no protocol/backend type enters example application code;
- no local Agent: GitHub-hosted Windows/Linux Qt 6.8.3 is sufficient for this slice.

## Deliverables

- `examples/widgets-basic` using `HyRemote::RemoteAccess` only;
- `examples/quick-basic` using the same facade against `QQuickWindow`;
- wire `HYREMOTE_BUILD_EXAMPLES` into the real build graph;
- deterministic test-mode options for loopback port, explicit remote input and bounded runtime without changing normal safe defaults;
- dual-OS GitHub Actions product-fit using maintained `vncdotool`;
- each example must prove actual framebuffer capture, remote pointer/button and keyboard delivery, disconnect/reconnect, clean stop and listener release.

## Acceptance boundary

This issue may close after the two examples and dual-OS standard-viewer E2E pass. It does **not** by itself close #6/#28/#29/#30: those parent issues still require the corresponding compatibility/documentation/local-coexistence matrix to be reviewed as a whole.

Governance mode: `transitional-explicit`.

skawu:
## Acceptance — E1/E2 public-facade dual-OS viewer E2E passed

[...]

Issue #41, opened 2026-09-15
[V1.0.0.0][EXAMPLES/DOCS] Deliver richer examples and user guides than Qt VNC Server


skawu:
## Product milestone

Supports **V1.0.0.0 GA** and all three mandatory HyRemote integration modes on x86_64 Windows + Linux.

Governance mode: `transitional-explicit` until #14 passes final ADS machine acceptance.

> **V1 convergence interpretation (2026-09-16):** all current implementation is on Draft PR #106. “Enable/disable remote control” in this acceptance issue means an explicit product policy, **not** a mandatory hidden live authorization channel. Embedded C++/QML may use the frozen explicit `stop -> configure -> start` lifecycle; zero-code QPA uses an explicit startup policy and relaunch. A live in-session toggle is not a V1 requirement unless the public API is deliberately expanded in a later release. This clarification preserves the minimal API/QPA design frozen by #101/#106.

## Owner requirement

HyRemote examples/documentation must be **richer, more detailed and more product-oriented than the two Qt VNC Server reference examples**.

Qt's official examples provide useful baselines:

- **Remote Desktop**: wraps a Qt Wayland Compositor output in `VncItem`, shares the desktop, and shows a simple connected-state indicator;
- **VncChat**: wraps a small QML subtree/TextArea in `VncItem`, demonstrates partial sharing, automatic port selection and connection/listening state.

HyRemote must go beyond these baselines by teaching all three product integration modes, Windows/Linux setup, deployment, security defaults, reconnect behavior, local+remote coexistence, troubleshooting and real external-consumer usage.

Examples are release acceptance artifacts, not demo-only extras.

### E1 — `examples/widgets-basic`

Purpose: minimum Embedded C++ API path for a normal Qt Widgets application.

Must demonstrate:

- `find_package(HyRemote)` / `HyRemote::RemoteAccess`;
- local window remains visible and interactive;
- remote view + pointer/keyboard input;
- explicit start/stop;
- view-only mode;
- reconnect;
- connection/session status displayed locally;
- Windows + Linux instructions.

### E2 — `examples/quick-basic`

Purpose: minimum Embedded C++ API path for a normal Qt Quick/QQuickWindow application.

Must demonstrate the same product semantics as E1, including local+remote simultaneous operation, input focus/text, resize/DPR and reconnect.

### E3 — `examples/qml-basic`

Purpose: Declarative QML API.

Target usage should remain compact, conceptually:

```qml
import HyRemote

RemoteAccess {
[...]

PR #61, opened 2026-09-15
feat(examples): add public-facade Widgets/Quick product E2E (#60)

skawu:
## Product context

Closes the largest remaining **V0.0.1.0 Embedded C++ product-evidence** gap under #6/#28/#29/#30 and supplies E1/E2 foundations for the V1.0 example program #41.

This PR is intentionally **stacked on #91 / PR #92** so E1/E2 can display truthful backend-neutral viewer status through `RemoteAccess::connectedClientCount()` rather than equating lifecycle `Running` with a connected client.

## Product examples

- `examples/widgets-basic`: normal Qt Widgets application using only `HyRemote::RemoteAccess`;
- `examples/quick-basic`: normal `QQuickWindow`/`QQuickView` application using the same facade;
- both keep construction inert and remote input view-only by default;
- both display `Listening` plus the real connected-client count locally;
- explicit `--remote-input` exists for the acceptance harness, not as a changed product default;
- no Core/Session/CaptureSource/InputSink/Transport/RFB types appear in application code;
- `HYREMOTE_BUILD_EXAMPLES` is connected to the real build graph.

## Product acceptance coverage

The Windows 2022 + Ubuntu 24.04 / exact Qt 6.8.3 workflow uses maintained `vncdotool` against the **actual examples**, not a transport-only test server.

Each Widgets and Quick example is exercised in two independent runs.

### Default view-only run

- starts without `--remote-input`;
- local diagnostic begins at client count 0;
- standard viewer connects and the local product diagnostic must change to 1;
- viewer receives a non-uniform framebuffer;
- viewer deliberately sends pointer/buttons, wheel and keyboard/text input;
- application must emit **no** `APP_POINTER`, `APP_WHEEL`, `APP_KEY` or `APP_TEXT` event;
- after viewer disconnect, local diagnostic must return to 0;
- public `RemoteAccess::stop()` remains clean and the listener no longer accepts afterwards.

This proves view permission and input permission are independent product policies rather than a UI label.

### Explicit control run

With `--remote-input`, the same viewer path verifies through the final Qt application adapter layer:

- real framebuffer capture from the attached target;
- left, middle and right pointer buttons;
- vertical wheel delivery;
- keyboard delivery;
- Shift/Ctrl/Alt modifier propagation into the resulting Qt key event;
- **text/IME commit independently from key delivery**;
- connected-client status sequence **0 → 1 → 0 → 1 → 0** across disconnect/reconnect without target-app restart;
- explicit stop and listener release.

The harness waits for the intermediate client-count 0 before reconnecting, so a broken diagnostic that collapses disconnect/reconnect cannot pass accidentally.

## Input-disconnect dependency

Static acceptance review found a separate #29 correctness gap: abrupt viewer disconnect while a key/button remains held can leave an unmatched press. **#90 / PR #93** now contains the private RFB fix plus raw abrupt-close product-fit. This examples PR does not duplicate that backend fix; #30 cannot claim the no-stuck-input requirement until #93 receives executable evidence and the composed product path is rerun.

## Boundaries

This PR introduces no new backend architecture and uses only the frozen public facade. Hosted/offscreen E2E proves viewer→shared runtime→Qt target behavior; it does not prove a physically visible local monitor/input path remains active. Final physical local+remote coexistence stays at #30/#33 acceptance.

Exact Windows/Linux execution remains blocked before runner assignment by #74, so this PR remains Draft and no dual-OS pass claim is made from failed no-step workflow runs.

Dependency: PR #92. Related correctness dependency: PR #93.


skawu (commented):
Owner code review — implementation shape is acceptable; keep Draft pending executable evidence.

Reviewed the public-example applications and E2E harness against the frozen V0.0.1 product boundary:
- both applications consume only `HyRemote::RemoteAccess` and keep remote input explicitly opt-in;
- Widgets and Quick use the same facade/lifecycle contract;
- the harness exercises a standard VNC client against the actual application targets, including framebuffer capture, pointer/key delivery, reconnect and listener release;
- no Core/transport/backend type leaks into application code.

Important acceptance boundary remains: the hosted E2E deliberately uses `QT_QPA_PLATFORM=offscreen` (and software Quick rendering), so it proves protocol→application correctness but **does not prove local-visible display/input coexistence**. That remains required at the parent #6/#28/#30 gate and must not be inferred from this CI.

Current GitHub Actions failures are infrastructure allocation failures (`steps=[]`, no runner assigned), not test failures. Do not merge until a real Windows + Linux run executes the workflow. No local Agent is required while repository work can continue in parallel.

Issue #62, opened 2026-09-15
[V0.0.3.0][QPA-01] Prove Qt 6.8.3 transparent native-delegate proxy skeleton


skawu:
## Parent

Child of #32. This is the first bounded implementation gate for the mandatory **Transparent QPA Proxy** path; it does not redefine V0.0.3.0 and does not weaken the local+remote coexistence requirement.

## Goal

For exact Qt 6.8.3, prove that a HyRemote QPA platform plugin can be selected as `QT_QPA_PLATFORM=hyremote`, instantiate the normal native platform integration as a delegate, and transparently forward the native local display/event services without exposing private-QPA types outside the dedicated adapter package.

Reference delegates for this gate:
- Windows x86_64: `qwindows`;
- Linux x86_64: `qxcb` under Xvfb for hosted CI (Xvfb is test infrastructure, not the product delegate).

## Frozen constraints

- this is a **proxy/decorator**, never Qt's replacement-only `qvnc` model;
- the native platform remains responsible for real local display, windows and native input;
- no QPA/private Qt type enters `hyremote-core`, `HyRemote::RemoteAccess`, or the QML API;
- target exact Qt 6.8.3 private ABI first; version coupling is explicit;
- no second Session/frame/input/transport semantics are invented here;
- do not claim remote capture/input from this first skeleton; that follows only after delegate preservation is proven;
- no local Agent while GitHub-hosted Windows/Linux can exercise the required x86 plugin load path.

## Implementation gate

- dedicated `qpa/` module and `HYREMOTE_WITH_QPA_PROXY` build path;
- QPA plugin metadata key `hyremote`;
- load the native delegate plugin without recursively selecting `hyremote`;
- forwarding `QPlatformIntegration` preserving capabilities, window/backing-store, event dispatcher and platform services;
- native delegate lifetime owned for the proxy lifetime;
- fail closed when delegate cannot be loaded (no silent fallback to headless/replacement display);
- dual-OS test app proves native delegate creation, window creation, event loop and orderly teardown.

## Security / launch contract for this gate

The native delegate is **not** accepted from arbitrary untrusted names. This first gate permits only the platform-native reference delegate (`windows` on Windows, `xcb` on Linux). The proxy rejects `hyremote`, `vnc`, `offscreen`, `minimal` and other replacement/headless delegates. This keeps the gate aligned with the local-display-preserving product requirement.

## Follow-on (not this gate)

[...]

Issue #32, opened 2026-09-15
[V0.0.3.0] Deliver mandatory Transparent QPA Proxy with local + remote coexistence on x86


skawu:
## Product milestone

**V0.0.3.0 — x86_64 (Windows + Linux) / Transparent QPA Proxy**

Governance mode: `transitional-explicit` until #14 passes reusable ADS machine acceptance.

> **V1 convergence authority update (2026-09-16):** the implementation candidate is now part of the single V1 line #106. This milestone authority is aligned with the frozen #101/#106 QPA contract: `qhyremote` is a package-owned exact-Qt platform MODULE, the application remains Qt-only, native delegate selection is HyRemote-owned for the qualified platform (`qwindows` / `qxcb`), and remote-input policy is a documented startup/relaunch policy rather than a hidden application runtime channel.

## Owner product decision

Transparent QPA Proxy is the **third mandatory HyRemote integration mode for V1.0.0.0 GA**. It is not an optional experiment.

HyRemote V1.0 must productize all three user-facing modes:

1. Embedded C++ API;
2. Declarative QML API;
3. Transparent QPA Proxy.

The QPA mode exists specifically for existing Qt applications that need **zero/minimal source modification while keeping native local display/input and adding remote view/control at the same time**.

## Important distinction from Qt's `qvnc` platform plugin

HyRemote QPA Proxy must **not** behave like a replacement-only remote platform backend.

The frozen architecture is a version-coupled **proxy/decorator platform integration**:

```text
Qt application
   ↓
HyRemote QPA Proxy (`qhyremote`)
   ├─ delegate normal platform operations → qualified native/local QPA backend
   │      Windows / Qt 6.8.3: qwindows
   │      Linux / Qt 6.8.3 reference path: qxcb
   │
   └─ compose/capture supported application surfaces → shared HyRemote RemoteAccess/Core/transport runtime

Remote input → HyRemote normalized input → qualified target/application surfaces
Local input  → native QPA delegate → target application
```

The application therefore continues to render locally through the native platform backend while HyRemote mirrors the supported application presentation path remotely.

This remains an acceptance claim only after exact-platform evidence passes; repository implementation alone does not make it Supported.

## User experience contract

A deployed ordinary Qt application enables the proxy without HyRemote application API calls:

```text
myapp -platform hyremote
```

Safe defaults are loopback listener, port 5900 and remote input disabled.

Explicit startup policy may be supplied with HyRemote-owned platform parameters, for example:

```text
[...]

PR #63, opened 2026-09-15
feat(qpa): prove exact-Qt native delegate proxy skeleton (#62)

skawu:
## Product context

First bounded implementation gate for **V0.0.3.0 — Transparent QPA Proxy** on the path to V1.0.0.0. This proves the prerequisite native-delegate-preserving QPA architecture; it does not claim the final remote-view/control product path.

## Scope

- opt-in exact Qt 6.8.3 private-ABI QPA package;
- plugin key `hyremote`;
- Windows delegate `windows`, Linux reference delegate `xcb`;
- native platform operations stay delegated to the real local backend;
- unsupported/headless replacement delegates are rejected;
- private QPA types stay isolated under `qpa/`;
- native window/event-loop smoke on both reference OS workflows.

## Acceptance boundary

QPA-01 alone is not #32 acceptance. Native-interface semantics, shared RemoteAccess composition, multi-surface capture/input, deployment, exact capture-family qualification, reconnect and physical local+remote coexistence remain in successor gates.

## Git Flow / evidence

This historical predecessor PR now targets `develop` under #95 while keeping its branch and stacked successor history intact. Hosted exact-Qt jobs remain affected by #74 and do not authorize merge/release/tag until they execute.


skawu:
Owner correction applied directly at head `3435b2f9`: Gate 01 smoke no longer treats `QWindow::visible`/Xvfb as local-visible product evidence. It now proves only the intended prerequisite: `hyremote` loads, the native delegate creates a platform window, the delegated event loop dispatches work, and teardown remains orderly.

Static Qt 6.8.3 review also confirms the current public virtual forwarding surface matches `QPlatformIntegration`, and `QPlatformIntegrationFactory::create()` explicitly loads the named native delegate. Two boundaries remain deliberately outside Gate 01 and must be qualified before #32 can close: (1) `QPlatformIntegration::call<>` uses dynamic-cast against the proxy object, so platform-native interface mixins are not automatically transparent; (2) protected keyboard-modifier/possible-key hooks cannot legally be forwarded through the base pointer and currently use base behavior. These are now a bounded follow-on, not hidden claims.

PR remains Draft until actual Windows/Linux execution occurs. Hosted jobs with no assigned steps are infrastructure evidence only and do not stop the next V1 work package.

skawu:
Branch/PR convergence cleanup: this QPA predecessor head is a verified ancestor of terminal convergence PR #97. It is closed unmerged as superseded; history/evidence remains preserved. Active QPA product convergence continues at #97, not on this micro-slice.

Issue #72, opened 2026-09-15
[V0.0.3.0][QPA-03] Auto-compose RemoteAccess behind the qualified native QPA proxy


skawu:
## Parent / dependency

Child of #32. **Follows #62 (QPA-01 native delegate skeleton) and #68 (QPA-02 native-semantics/interception qualification).** This gate begins the zero-application-code remote-access composition required by Transparent QPA Proxy mode; it does not bypass or replace #68 and does not by itself close V0.0.3.0.

Implementation may proceed in a stacked branch while #68 is being qualified, but this gate cannot be accepted/merged as product evidence before the required #68 native-semantics boundary is accepted.

## Goal

Selecting `QT_QPA_PLATFORM=hyremote` must preserve the qualified native `qwindows` / `qxcb` local path and, once a supported visible top-level Qt target exists, automatically compose the **same `HyRemote::RemoteAccess` runtime** used by Embedded C++ and QML modes. No second Session/transport/input semantics are allowed.

## Frozen launch contract

- selecting the `hyremote` platform plugin is the explicit act that enables the QPA remote-access mode;
- no network listener is opened merely by loading the plugin;
- composition begins only after a supported visible application target exists and the qualified native delegate path remains authoritative;
- default listen address: `127.0.0.1`;
- default port: `5900`;
- remote input remains **disabled by default**;
- QPA-specific parameters may override address/port/input, but invalid values fail closed and must not silently expose a wildcard listener;
- native platform-specific parameters not owned by HyRemote continue to the native delegate.

## Gate-03 target selection

For this bounded gate, automatically select one primary supported visible top-level application target:
- a visible top-level `QWidget` when running a `QApplication` and the Widgets adapter is built;
- otherwise a visible top-level `QWindow` that the normal `HyRemote::RemoteAccess` target adapters accept (for example `QQuickWindow`).

The controller must stop/re-evaluate when that target is destroyed/hidden. **Multi-window composition, popup/dialog aggregation and full application-surface semantics are follow-on #76 work and are not to be misrepresented as complete here.**

## Architecture constraints

- #68 remains authoritative for native-interface/protected-hook/interception semantics;
- QPA/private Qt types remain under `qpa/`;
- the automatic controller consumes only the public `HyRemote::RemoteAccess` facade for this single-target gate;
- Core remains Qt/QPA/protocol free;
- QML and explicit C++ application behavior is unchanged;
[...]

Issue #68, opened 2026-09-15
[V0.0.3.0][QPA-02] Qualify native QPA semantics and add an interception-safe proxy seam


skawu:
## Parent / dependency

Child of #32. Follows #62 / PR #63 Gate 01.

Gate 01 proves the bounded prerequisite only: `QT_QPA_PLATFORM=hyremote` can create the reference native platform delegate (`windows` / `xcb`), obtain a platform-window path, keep the delegated event loop alive and tear down without replacing the product with a headless/qvnc-style backend.

This issue owns the next architectural gate before any QPA remote-capture product claim.

## Goal

For exact Qt 6.8.3 on Windows x86_64 and Linux x86_64, prove that the proxy preserves the native-QPA semantics required by ordinary Qt applications **and** introduce a narrow interception seam suitable for later presentation mirroring without moving QPA/private types into Core or duplicating `RemoteAccess` semantics.

## Why this gate is mandatory

A generic `QPlatformIntegration` forwarding wrapper is not automatically transparent for all Qt private/native behavior:

- `QPlatformIntegration::call<>` performs a `dynamic_cast` against the integration object itself. Native-interface mixins implemented by `qwindows`/`qxcb` therefore do not become visible merely because `nativeInterface()` is forwarded.
- protected hooks such as keyboard-modifier and possible-key queries cannot legally be invoked through an arbitrary base pointer; Gate 01 currently uses the Qt base behavior for these hooks.
- final V0.0.3.0 requires local native behavior to remain authoritative while HyRemote adds remote view/control. We must not discover a native-semantic regression only after capture/input composition is added.

## Frozen design constraints

- QPA Proxy remains a **native-delegate-preserving proxy/decorator**, never a replacement-only `qvnc` design;
- exact Qt 6.8.3 private ABI only for this gate;
- Windows reference delegate remains `windows`, Linux remains `xcb`;
- no QPA/private Qt type enters `hyremote-core`, `HyRemote::RemoteAccess`, or the QML module;
- no duplicate Session/capture/input/transport semantics;
- no attempt to make one generic private-QPA implementation promise compatibility across unrelated Qt versions;
- no remote framebuffer/product claim until native semantics and the interception boundary are proven;
- hosted x86 validation is used where sufficient; no Local Agent while repository/hosted evidence can prove the slice.

### A — native-semantics inventory and qualification

[...]

PR #75, opened 2026-09-15
feat(qpa): auto-compose the shared RemoteAccess runtime after native qualification (#72)

skawu:
## Product context

Stacked implementation for **#72 / QPA-03 — automatic RemoteAccess composition**. The canonical predecessor order is now explicit:

`#62 QPA-01 native delegate skeleton → #68 QPA-02 native-semantics/interception qualification → #72 QPA-03 this gate → #76 QPA-04 multi-surface composition`.

This PR may continue implementation while #68 is being developed, but it **must not be accepted/merged as product evidence before #68 passes**.

## What changes

- QPA mode requires and links the same `HyRemote::RemoteAccess` runtime used by Embedded C++ and QML modes;
- loading the platform plugin itself still opens no listener;
- after a real native platform window is created, a private controller is armed and attaches to a supported visible top-level application target;
- Widgets applications prefer their real top-level `QWidget`; other applications fall through to visible top-level `QWindow` targets accepted by normal adapters (for example `QQuickWindow`);
- target hide/destruction stops the current remote runtime and triggers target re-evaluation;
- no QPA/private Qt type enters Core, RemoteAccess public API or QML;
- QPA enablement unconditionally qualifies the exact Qt 6.8.3 package/private ABI.

## Safe launch contract

Defaults remain loopback / port 5900 / remote input disabled. QPA-specific parameters are explicit:
- `hyremote-address=<IP>`;
- `hyremote-port=<1..65535>`;
- `hyremote-input=on|off` (also 1/0, true/false, yes/no).

Invalid HyRemote parameter values fail closed. Non-HyRemote platform parameters are preserved for the native delegate. An explicit wildcard address is possible only when actually supplied by the user.

## Lifecycle correction

The RemoteController is **not** started from `QPlatformIntegration::initialize()`. It is armed lazily after the native delegate successfully creates the first platform window, avoiding an early-application-lifecycle race while preserving the rule that loading the plugin alone does not listen.

## Tests / CI

- deterministic config/default/fail-closed coverage;
- normal QWidget QPA smoke expects an RFB 3.8 listener without application HyRemote code, then hides the target and verifies listener release;
- Windows/Linux QPA workflow enables Core + RemoteAccess + Widgets/Quick + bounded RFB and runs all `hyremote-qpa-*` tests.

## Acceptance boundary

This remains **single-primary-target QPA-03**. Multi-window/popup/dialog composition is #76. Native-interface/protected-hook/interception semantics remain #68 and are a hard predecessor. Hosted execution is required before merge; current no-runner GitHub Actions failures are not pass evidence.


skawu:
Superseded by PR #79.

Reason: this PR was implemented before #68/QPA-02 existed and remained based directly on the QPA-01 branch. #79 recreates the same bounded QPA-03 capability **on top of the actual #68 implementation**, preserving its exact Qt 6.8.3 native-interface, key-mapper and ownership-safe interception corrections.

No product requirement is dropped; this is dependency-DAG convergence so QPA-03 cannot structurally bypass its hard predecessor.

Issue #76, opened 2026-09-15
[V0.0.3.0][QPA-04] Compose multi-window application surfaces without reconnecting the viewer


skawu:
## Parent / dependency

Child of #32. Follows the ordered QPA chain:

1. #62 — QPA-01 native delegate proxy skeleton;
2. #68 — QPA-02 native-semantics/interception qualification;
3. #72 — QPA-03 automatic single-primary-target RemoteAccess composition;
4. **this issue — QPA-04 multi-surface application composition.**

Implementation of the deterministic surface model may proceed in a stacked branch, but product acceptance cannot skip the predecessor gates.

## Product requirement

A normal Qt application using `QT_QPA_PLATFORM=hyremote` must remain one remote application session when it opens/closes/moves independent top-level application surfaces. A modal dialog, menu/tool popup or second application window must **not** require the VNC viewer to disconnect/reconnect merely because the active application surface changed.

HyRemote remains an **application remote-access framework**, not an arbitrary OS desktop server. Only surfaces owned by the current Qt application are candidates for composition.

## Frozen behavior

- preserve the qualified native `qwindows` / `qxcb` local display and native input path from #68;
- keep one bounded Core Session / one transport listener for the application remote session;
- maintain a logical application canvas composed from eligible visible top-level application surfaces;
- track surface geometry, visibility, activation/stacking intent and lifetime;
- update framebuffer geometry through the existing RFB desktop-size path when the application canvas changes;
- remote pointer coordinates are mapped back to the topmost eligible application surface at that point;
- keyboard/text events go to the current application focus/active target rather than inventing protocol-specific focus semantics;
- opening/closing a dialog/popup/window must not restart the transport solely to switch targets;
- local interaction continues normally and is not captured/replaced by a synthetic desktop environment.

## Surface scope for this gate

At minimum qualify:
- primary QWidget top-level window;
- independent `QDialog`/tool window;
- common QWidget popup/menu behavior where Qt exposes a capturable application surface;
- `QQuickWindow` primary/secondary top-level windows;
- Qt Quick in-window Popup/ToolTip continues through the normal scene capture path and is not double-composed.

Native/foreign windows that cannot be captured with the qualified Qt-facing path must be reported as a limitation rather than silently claimed.

## Architecture constraint

Do **not** solve this by creating one public `RemoteAccess` per window or one VNC listener per surface. Do not leak QPA/private Qt types into Core or the public C++/QML API.

[...]

PR #77, opened 2026-09-15
feat(qpa): add deterministic QPA-04 application surface model (#76)

skawu:
## Product context

First bounded implementation slice of **#76 / QPA-04 multi-surface application composition**. Ordered predecessor chain:

`#62 QPA-01 → #68 QPA-02 native semantics/interception → #72 QPA-03 automatic single target → #76 QPA-04 multi-surface`.

The model can be implemented/reviewed in parallel, but this PR cannot be accepted as final product evidence by skipping its predecessors.

## Scope

Adds a QPA-private `ApplicationSurfaceModel` defining:
- visible application canvas as the union of eligible top-level surface geometries;
- native/global coordinates translated into one remote-canvas coordinate space, including negative screen coordinates;
- deterministic back-to-front ordering;
- explicit raise/show semantics;
- topmost-surface pointer routing with local-coordinate conversion;
- gaps in the rectangular canvas remaining non-interactive;
- hide/remove/move updates without implying a transport restart.

## Architecture boundary

The model depends only on QtCore geometry/value types. It knows nothing about RFB, Core Session, capture storage, QPA private classes or protocol input. No public API changes are introduced.

This PR intentionally does **not** yet implement frame compositing or a multi-surface input sink. Those follow after this deterministic model is accepted; final #76 must keep one transport session when windows/dialogs change and must build on the qualified #68 interception seam rather than bypass native semantics.

## Deterministic coverage

Pins negative coordinates, overlap/topmost routing, explicit raise, canvas expansion, non-interactive gaps, hide/show/remove behavior, and geometry churn preserving z-order.

Hosted execution is required before merge. Current GitHub hosted-runner allocation failures are not pass evidence and are not a reason to stop repository implementation work.


skawu:
Superseded by PR #80.

#80 carries the same deterministic application-surface model slice but is stacked on the corrected chain `#68 / PR #78 → #72 / PR #79`. The old PR #77 remained based on the pre-#68 QPA-03 branch. No model/product requirement is dropped; this is dependency-DAG convergence before runtime multi-surface composition continues.

Issue #90, opened 2026-09-15
[V0.0.1.0][INPUT-RESET] Release held remote input state on viewer disconnect


skawu:
## Parent / goal

Child of #29 and blocking evidence for #30 / V1.0.0.0.

The current bounded RFB transport stores each viewer's `buttonMask` and modifier state inside `ClientState`, but the disconnect path erases that state without emitting balancing release events. A viewer that disconnects while a mouse button or modifier key is held can therefore leave the Qt target with an unmatched press.

This directly violates #29's acceptance requirement that key-up state is not left stuck after disconnect.

## Required behavior

For a client that reached the normal RFB input phase:

- track enough per-client pressed input state to balance events on disconnect;
- before discarding the client state, emit release events for held pointer buttons and held keys/modifiers through the same normalized `hyremote::InputEvent` path;
- do not inject synthetic releases for a client that never reached the normal/announced state;
- disconnect cleanup must remain bounded and must not block the Qt/network event loop;
- normal key/button release messages must remove the tracked state so disconnect does not duplicate them;
- one viewer's disconnect cleanup must not mutate another viewer's input state;
- stopping the transport must not publish late input after the input callback has been disabled.

## Scope boundary

This is a transport normalization/correctness fix, not an OS-level virtual-input mechanism and not a new public API. Widgets and Quick continue consuming the same normalized InputSink contract.

No protocol/backend types may leak into Core or `HyRemote::RemoteAccess`.

## Evidence

Add deterministic transport evidence for at least:

1. Shift/key press followed by abrupt viewer disconnect produces corresponding normalized key release(s);
2. held left/middle/right button followed by disconnect produces releases at the last known pointer location;
3. already released input is not released twice;
4. a subsequent viewer starts with clean modifier/button state;
5. product E1/E2 viewer acceptance can exercise abrupt disconnect without leaving Widgets/Quick input state stuck.

Hosted Windows/Linux execution remains subject to #74; repository implementation and deterministic test construction must continue without substituting Local Agent for hosted infrastructure.

Governance mode: `transitional-explicit`.

skawu:
Implementation is now in Draft PR #93.

Repository evidence constructed:
- private RFB per-client held-key / modifier / pointer-button cleanup;
- bounded 64-key tracking and immediate protocol-failure exit on overflow;
- abrupt raw-RFB disconnect product-fit proving balancing A/Shift/left/right releases, no duplicate middle release, and clean subsequent-client modifier state;
[...]

Issue #74, opened 2026-09-15
[V1.0.0.0][CI-INFRA] Restore Windows/Linux hosted runner execution for acceptance evidence


skawu:
## Product impact

This is a release-evidence infrastructure blocker for the continuous V1.0.0.0 path. It does **not** stop repository implementation work, but it prevents the project from honestly closing dual-OS acceptance gates that require executable CI evidence.

Affected product gates include #11, #30, #31, #32, #33 and dependent example/package/QPA tasks.

## Observed failure mode

Multiple independent workflows on current V1-path branches terminate almost immediately before any job step is assigned.

Representative evidence:

- PR #61 / commit `5b3d76b22aa4762de564fa44a75a6015c696378b`
- workflow: `Embedded C++ product examples E2E`, run `34957162373`, attempt 2
- Windows job `104342233686`: `steps=[]`, `runner_id=0`, empty runner name, completed failure in ~2 seconds
- Linux job `104342233977`: same signature
- PR #51 declarative workflow run `34956996957` shows the same no-runner/no-step signature on both OSes
- unrelated RemoteAccess/SDK/Widgets/Quick/RFB workflows fail in the same way, so this is not presently attributable to one workflow's build/test commands.

## Required disposition

Determine and restore the repository/account/Actions state required for GitHub-hosted `windows-2022` and `ubuntu-24.04` jobs to receive runners and execute steps.

Possible external causes (billing/quota/account policy/GitHub Actions availability) must be verified from the actual repository/account state rather than guessed in code.

## Evidence after restoration

At minimum rerun the current V1 critical-path workflows for:

- production RFB transport;
- RemoteAccess facade + Widgets/Quick adapters;
- Embedded C++ public examples E2E (#61);
- Declarative QML API / example / deployment (#51/#67/#70);
- QPA native-delegate gates (#63 and follow-ons);
- remote-support-showcase (#73 after its stacked base is available).

A successful runner assignment is necessary but not sufficient: each product gate still requires its own workflow/test acceptance.

## Execution rule

Repository work continues while this issue is open. Do not replace missing Windows/Linux evidence with inferred claims, and do not use Local Agent merely to compensate for a hosted-runner/account infrastructure problem unless a separate acceptance task genuinely requires local display/hardware evidence.

Governance mode: `transitional-explicit`.

skawu:
Additional independent reproduction from QPA-02 / PR #78:

[...]

PR #106, opened 2026-09-16
ci(v1): add integrated GA acceptance matrix (#104)

skawu:
## V1.0.0.0 single convergence authority

This Draft PR is the **only active V1 product-integration line**. Earlier implementation slices are historical audit/evidence inputs only. It targets `develop`; it cannot authorize a release/tag by itself.

Current reviewed head: `eab386ef8d41e618c11b94166b7847fb9d2bdb0f`. All 9 current-head hosted workflows have actually executed and are green, but this SHA is **not final** because V1-targeting PRs and mandatory correctness/scope work remain open.

Current handoff/continuation authority: #156. Systemic retrospective: #161. Candidate freeze/change control: #165.

## Frozen product/application architecture

V1 retains exactly three first-class modes:

### Embedded C++

```cmake
find_package(HyRemote CONFIG REQUIRED)
target_link_libraries(MyApp PRIVATE HyRemote::RemoteAccess)
```

```cpp
#include <HyRemote/RemoteAccess.h>
HyRemote::RemoteAccess remote(&window);
remote.start();
```

### Declarative QML

```qml
import HyRemote
RemoteAccess {
    target: mainWindow
    enabled: true
}
```

Thin wrapper over the same normal shared runtime; no second Core/Session/transport stack.

### Transparent QPA

```text
MyApp -platform hyremote
```

```cmake
find_package(HyRemote CONFIG REQUIRED)
hyremote_deploy(TARGET MyApp QPA)
```

Ordinary applications remain Qt-only. `qhyremote` is a native-delegate-preserving exact-private-ABI platform payload, not an application link target. The V1 reference QPA line remains **Qt 6.8.3 exact** with `qwindows` / `qxcb`.

## Frozen artifact/package model

- one normal application C++ target/runtime: `HyRemote::RemoteAccess` / shared `HyRemoteRemoteAccess`;
- `hyremote-core` stays STATIC internal/source composition and is not installed/exported as a normal application target;
- QML is package payload over the same runtime;
- `qhyremote` is an exact-Qt MODULE payload and is not exported as `HyRemote::QpaPlatform`;
- `BUILD_SHARED_LIBS` must not create alternate product personalities;
- one deployment entry point: `hyremote_deploy(TARGET ... [QML] [QPA])`;
- source and installed SDK consumption preserve the same application contract;
- optional QML/QPA selection remains fail-closed.

## Repository/layout boundary

[...]

PR #113, opened 2026-09-17
fix(input): release held remote input when a viewer disconnects (#90)

skawu:
## Why (issue #90)

The bounded RFB transport kept each viewer's `buttonMask`/modifier state inside `ClientState`, but the
disconnect path erased that state **without emitting balancing releases**. A viewer that disconnected
while holding a mouse button or a modifier key therefore left the Qt target with an unmatched press,
which is exactly what #29's acceptance forbids.

## What changed

| File | Change |
|---|---|
`remoteaccess/src/transport/rfb_transport.cpp` | `ClientState` now also tracks the last reported pointer location/viewport and a fixed-size (cap 16) array of held `KeyCode`s; new `releaseHeldInput()` emits one pointer-button release per still-set bit at the last known location and key releases in reverse press order with the modifier set shrinking; `rememberHeldKey()`/`forgetHeldKey()` keep the tracking correct on normal presses/releases; the pointer button table is now a shared `constexpr` used by both the pointer handler and the release path |
`remoteaccess/tests/test_rfb_input_release.cpp` (new) | Deterministic loopback evidence: a minimal RFB 3.8 client drives the real transport |
`remoteaccess/tests/CMakeLists.txt` | Registers `hyremote-rfb-input-release-test` (inside the existing `HYREMOTE_WITH_VNC` guard) |

Design points that follow the issue text:

- **Bounded and non-blocking**: at most one event per held button/key, no allocation, no waiting on the
  Qt/network event loop thread.
- **Only announced viewers**: releases are emitted only for a client that reached the input phase; a
  viewer that never got there (case 4) receives nothing synthetic.
- **No duplicates**: a normal release removes the tracked state, so the disconnect path has nothing left
  to balance (case 3).
- **No cross-viewer effects**: only the disconnecting viewer's own state is read or mutated (case 5).
- **No late input on stop**: `publishInput` already suppresses events once the transport is stopping,
  and `shutdown()` disconnects the socket handlers before aborting them (case 6).
- **Scope boundary respected**: no public API change, no protocol/backend types in Core or
  `HyRemote::RemoteAccess`; Widgets/Quick keep consuming the same normalized `InputSink` contract.

## Evidence

```
build/all/remoteaccess/tests/hyremote-rfb-input-release-test.exe
  -> rfb input-release evidence: 6 case(s), 0 failure(s)      (EXIT=0)

ctest --test-dir build/all
  -> 100% tests passed, 0 tests failed out of 30
```

**Control experiment** (the project's practice of proving a new test can fail): with the
`releaseHeldInput(*it->second);` call commented out, the same binary reports **8 assertion failures**
confined to the disconnect cases (held key release, button releases at the last location, cross-viewer
independence), while the no-duplicate / unannounced-viewer / stop-silence cases still pass - the exact
pre-#90 behaviour. The control was then reverted and the suite is green again; no control markers remain
in the tree.

Cases map to the evidence list in the issue: 1 -> case 1, 2 -> case 2, 3 -> case 3, 4 -> case 5.

## Not covered by this PR

- **Issue evidence item 5** (product E1/E2 viewer acceptance exercising abrupt disconnect) depends on
  the `examples/` E1/E2 applications, which live on the V1 convergence line, not on `main`.
- **Hosted Windows/Linux execution** of this new test stays subject to #74; the evidence above is the
  local reference environment (Windows 11 x86_64, MSVC 14.44, Qt 6.8.3), and the test is registered with
  CTest so the hosted matrix picks it up once #74 is unblocked.


skawu:
Superseded by the single V1 convergence line in Draft PR #106. The current #106 tree contains the newer bounded-RFB disconnect model under the canonical `src/remoteaccess` layout, including per-viewer held-state accounting, shared logical key/button holder reference counts, unmatched-release isolation, explicit disconnect cleanup, protected-release mailbox admission under backpressure, and deterministic real-RFB multi-client regressions. Keeping this older `main`-based implementation open would create a second input-authority path and risks reintroducing the old repository layout/semantics. Historical evidence in this PR remains useful, but no V1 development should continue here; #106 is the implementation authority.

PR #116, opened 2026-09-17
chore(ci): cancel superseded runs and bound job duration

skawu:
## Why

Every workflow in this repository started a fresh full matrix on every push and never cancelled a
superseded run. A branch that receives commits faster than the matrix completes therefore fills the queue,
and the backlog itself becomes the evidence bottleneck rather than the runner supply:

* 2026-09-17: **308 runs queued** with only 18 in progress (`GET /actions/runs?status=queued`), while the
  account-level denial was already over and githubstatus reported Actions `operational` - i.e. runners were
  being granted, the queue was the problem.
* inflow ~17-19 runs per push (most workflows define 3 job blocks plus a matrix) versus ~36 completed
  jobs/hour at ~18 concurrent slots and 16-60 minute jobs.
* as a result, by the time a head's jobs were actually scheduled that head was 20+ commits old (issue #74).

## What

For each of the seven workflows defined on `main`:

* add a per-workflow, per-ref concurrency group with `cancel-in-progress: true`, so a newer push of the same
  workflow on the same ref cancels the superseded run instead of queueing behind it;
* bound every job with `timeout-minutes: 90` (one observed job ran 62 minutes).

These seven files are also the ones inherited by PR #106 - its branch only *adds*
`v1-ga-acceptance.yml`, `qml-api.yml`, `qpa-proxy.yml` and `git-flow-policy.yml` - so this change reduces the
per-push fan-out on that PR as well.

## Verification

* `git diff --stat`: 7 files changed, 35 insertions, no change to build or test behaviour.
* YAML-only change validated by the next `pull_request` run on this branch: for a `pull_request` event the
  workflow definitions come from the merge commit, so this PR exercises its own new `concurrency` block.
* Baseline measured before the change: `queued = 308`, `in_progress = 18`, oldest queued ~40 minutes.

## Related

* Issue #74 - this removes the self-inflicted half of the evidence problem; the remainder is runner supply.
* The four workflows that exist only on the V1 convergence branch still have no concurrency group and need
  the same three lines (`v1-ga-acceptance.yml`, `qml-api.yml`, `qpa-proxy.yml`, `git-flow-policy.yml`). I have
  not modified that branch.


skawu:
V1 convergence decision: the queue-control part of this PR has been superseded by #117, which was merged directly into the single active V1 convergence branch #106. All nine V1 workflows now use per-workflow/per-ref `concurrency` with `cancel-in-progress: true`, so the self-inflicted backlog is addressed at the actual candidate source.

The remaining proposal here is the blanket `timeout-minutes: 90` policy. I am deliberately not making that a V1 release requirement yet: hosted runner assignment/execution has only just resumed intermittently, and adding a new universal timeout before the current all-modes jobs have stable timing evidence could turn infrastructure variability into an artificial product failure. The new post-merge mainline validation added on #106 is bounded at 90 minutes, but the V1 candidate workflows will keep their existing semantics until real current-head runs show a concrete runaway-job problem.

No product/runtime/API content remains unique to this PR. Closing it as superseded by #106/#117; the PR remains the audit record for the queue analysis.

Issue #104, opened 2026-09-16
[V1.0.0.0][GA-AUTOMATION] Unified Windows/Linux acceptance matrix for all three integration modes


skawu:
## Parent / release authority

Child of #33. This issue owns the **single integrated GA automation gate** for HyRemote V1.0.0.0. It does not replace milestone authorities #30, #31, #32, #39 or #41; it composes the accepted product surfaces into one exact-candidate Windows/Linux matrix.

> **Current state (2026-09-16):** `.github/workflows/v1-ga-acceptance.yml` and the integrated release-like candidate are implemented on Draft PR #106 (`feature/104-v1-ga-acceptance-matrix -> develop`), which is now the only active V1 product-integration line. New V1 fixes discovered by this gate go to #106; do not create another GA convergence branch.

Governance mode: `transitional-explicit` until #14 passes reusable ADS machine acceptance.

## Why this gate exists

Focused Core, RemoteAccess, Widgets, Quick, QML, SDK, QPA and example tests can each pass while the assembled release candidate still fails because of target graph, package, runtime, deployment or cross-mode interaction. V1 therefore requires one canonical all-modes run over the same tree and installed product shape.

The gate must never:

- turn #74 no-runner failures into code results;
- substitute hosted/headless/Xvfb execution for #109 physical/native local-display/input evidence;
- broaden the frozen public/package architecture merely to make CI easier;
- infer support for another Qt/OS combination from the reference run.

### 1. Embedded C++ API

- E1 Widgets and E2 Quick;
- application links only `HyRemote::RemoteAccess`;
- one shared `HyRemoteRemoteAccess` runtime;
- view-only safe default and explicit remote-input path;
- active view/control policy changes follow `stop -> configure -> start` rather than a hidden live channel;
- framebuffer/input/connect/disconnect/reconnect/client-count behavior.

### 2. Declarative QML API

- E3 `import HyRemote` / QML `RemoteAccess`;
- thin wrapper over the same shared runtime;
- component-complete initial enablement;
- stopped-runtime configuration semantics shared with C++;
- view/input/reconnect/error/client-count parity;
- no duplicate Core/transport/session implementation.

### 3. Transparent QPA Proxy

- exact Qt **6.8.3** private-ABI gate;
- package-owned `qhyremote` MODULE;
- ordinary E4 application remains Qt-only;
- native delegate preservation (`qwindows` / `qxcb` reference path);
- one application-level shared runtime/listener across qualified surface churn;
[...]

PR #120, opened 2026-09-17
build: restore buildability of the frozen V1 candidate after the head stopped configuring

skawu:
## What this is

This is option (A) from my adjudication comment on PR #106 (comment `5712434396`): the current head
`0535f87` cannot configure at all, so the convergence line currently produces no buildable evidence and the
#104/#109 entry conditions are unreachable. These four commits restore buildability. Nothing here changes a
product behaviour, an acceptance cell, a compatibility statement or a workflow.

Each commit is a separate defect with its own evidence. Merge, cherry-pick individually or discard - all
yours.

## 1. `3f35a34` build: restore the release-profile validation call so the candidate configures again

`CMakeLists.txt:17` called `hyremote_apply_release_profile()`, which is defined nowhere on this branch
(`cmake/HyRemoteReleaseProfile.cmake` defines `hyremote_validate_release_profile`, a validator). Every job
that configures the project failed, e.g. job `105131974942`:

```
CMake Error at CMakeLists.txt:17 (hyremote_apply_release_profile):
  Unknown CMake command "hyremote_apply_release_profile".
-- Configuring incomplete, errors occurred!
```

The commit restores the working form and the pieces that went with it: the
`hyremote_validate_release_profile(VERSION ... QML_ENABLED ... QPA_ENABLED ...)` call, the `HyRemoteDeploy`
include (install is now included only when the product target exists), the QML import-path reset and
`HYREMOTE_QML_SOURCE_DEPLOY_TARGETS` property, the `HYREMOTE_BUILD_TESTS`-guarded `include(CTest)`, and the
configuration status output. It also keeps the option-guarded `add_subdirectory` calls with explicit
diagnostics, so a misconfigured consumer gets a message instead of a silent half-build.

It additionally registers the milestone release-profile matrix that the validator's own comment describes, so
the documented milestone -> integration-mode mapping is enforced rather than implied: 0.0.0 (sentinel) all
modes, 0.0.1.0 C++ only, 0.0.1.0 rejects QML, 0.0.1.0 rejects QPA, 0.0.2.0 QML, 0.0.2.0 rejects QPA,
0.0.3.0 all modes, and 1.0.0.0 all modes. The negative cases use `WILL_FAIL`, so they assert the guard
actually rejects.

## 2. `3d0926f` build(qpa): use the Qt 6.8.3 QPA header path and include QPalette

Two compile defects in `integrations/qpa/hyremote_platform_plugin.cpp`:

- line 9 included `<QtGui/private/qplatformintegrationfactory_p.h>`. The exact qualified Qt 6.8.3 kit ships
  that header as `include/QtGui/6.8.3/QtGui/qpa/qplatformintegrationfactory_p.h`, i.e. in the QPA header set,
  not under `private/`. The include is now the `qpa/` form already used by the four neighbouring includes in
  the same file.
- the file overrides `populateLightSystemPalette(QPalette &)` but never included `QPalette`, so the compile
  failed with `error C2664 ... undefined type QPalette` at line 272 even after the header path was fixed.

## 3. `3bfacfd` build(qpa): give the composite tests the RemoteAccess include surface they compile against

`hyremote-qpa-composite-capture-test` and `hyremote-qpa-composite-input-test` compile sources that include
the RemoteAccess private component-factory declarations, which reference `QHostAddress` and the generated
`HyRemote/RemoteAccessExport.h`. Neither target had those include directories. Both now link
`HyRemote::RemoteAccess`, exactly as the neighbouring `hyremote-qpa-remote-config-test` already does, which
supplies the public include directory, the generated header directory and Qt6::Network transitively.

## 4. `0c6227d` build(examples): enable AUTOMOC before qt_add_qml_module in the QML example

`examples/qml-basic` could not even be configured. Control experiment (AUTOMOC line removed, everything else
identical):

```
CMake Error at C:/Qt/.../Qt6QmlMacros.cmake:3689 (qt6_extract_metatypes)
Call Stack: ... qt6_add_qml_module -> examples/qml-basic/CMakeLists.txt:5 (qt_add_qml_module)
-- Configuring incomplete, errors occurred!
```

The target therefore sets `CMAKE_AUTOMOC` before `qt_add_executable`, the same rule
`integrations/qml/HyRemote` already follows for the module itself.

## Verification (local, exact qualified Qt 6.8.3 / MSVC x64, clean `build/ga` tree)

```
cmake -S . -B build/ga -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_PREFIX_PATH=<Qt 6.8.3 msvc2022_64> \
      -DHYREMOTE_BUILD_TESTS=ON -DHYREMOTE_BUILD_EXAMPLES=ON \
      -DHYREMOTE_BUILD_QML_API=ON -DHYREMOTE_WITH_QPA_PROXY=ON
cmake --build build/ga -- -k 0
ctest --test-dir build/ga            # with the build's remoteaccess/qml/qpa directories on PATH
```

- configure: `Configuring done` / `Generating done` (before: the `Unknown CMake command` error above).
- build: every target links **except** `hyremote-qpa-platform` (see the blocker below).
- tests: **60/63 pass**, including all core, remoteaccess, RFB, QML-module, deployment-helper and
  release-profile cases. The three that do not pass:
  - `hyremote-release-readiness-metadata` - pre-existing and unrelated to this change: the gate rejects the
    development sentinel version (`root project(VERSION ...) is missing or malformed`,
    `tests/release-readiness/check_release_metadata.cmake:78`). It fails the same way without this branch.
  - `hyremote-qpa-proxy-smoke` and `hyremote-qpa-native-semantics` - they need the platform plugin, which
    cannot be built until the blocker below is resolved.

## The one remaining blocker is a design question, not a link flag

`hyremote-qpa-platform` calls three symbols that live in the Core static library:

- `hyremote::CpuFrameStorage::createSinglePlane(...)` and `hyremote::CpuFrameStorage::mutablePlane(...)`
  (from `composite_target.cpp`),
- `hyremote::mapPointerToTarget(...)` (from `interactive_composite_target.cpp`).

Adding `HyRemote::Core` to that target does make everything link, and with it the suite reaches **61/63 with
both QPA smoke tests passing** - but the repository's own gate rejects that link:

```
repository-layout: QPA platform payload must not link Core directly: HyRemote::Core
  tests/release-readiness/check_repository_layout.cmake:40 (forbid_link_target)
```

So I reverted it and left the link line exactly as committed: the payload has to reach those capabilities
through the runtime instead. That is your call, not mine to make silently. Two shapes are available:

1. `HyRemote::RemoteAccess` exposes the frame-storage allocation and the input mapping the composition needs
   (a runtime-side addition, must respect the frozen public API contract), or
2. the composite composition stops touching Core and asks the runtime to publish frames and route input.

Both are behaviour-preserving by intent, and either is a small PR once you choose the shape. I have not
touched `src/**`, the release-readiness gates, the GA matrix, the workflows or the #109 exclusion statements.

## Local evidence, for reproduction

- Number of failed suite members at head: 3 (1 pre-existing gate, 2 blocked by the plugin).
- With the forbidden `HyRemote::Core` link added as an experiment: build complete, 61/63, QPA smoke tests
  green - recorded here as evidence of the trade-off, not as a proposal.


skawu:
## Narrowed to the authorized shape, and the P1 from the automated review is resolved

Per the #106 ruling (option A, surgical repair of the accidental root rewrite), this PR now touches **root
`CMakeLists.txt` only** - one file, one commit. The three extra files from the previous revision (the QPA
plugin includes, the QPA composite-test link surface, the QML example AUTOMOC) have been removed from the
branch and are listed below as the downstream queue.

What the single commit does, exactly as ruled:

- restores the known-good root `CMakeLists.txt` from `99215c3`, the parent of the accidental `0535f87`;
- keeps the one intended semantic change from #111 - the description stays `Qt Remote Access Framework`;
- restores `hyremote_validate_release_profile(VERSION ... QML_ENABLED ... QPA_ENABLED ...)` and does **not**
  invent `hyremote_apply_release_profile()`;
- restores the `HyRemoteDeploy` include and the QML source-deployment metadata setup
  (`HyRemote_QML_IMPORT_PATH` reset, `HYREMOTE_QML_SOURCE_DEPLOY_TARGETS`), the target/Qt availability checks,
  top-level/subdirectory behaviour, the release-readiness test registrations (including the eight-case
  milestone release-profile matrix) and the install sequencing.

Two deltas beyond the raw restore, both inside the same file and both backed by local evidence:

1. **The `project()` declaration is on one line.** The restored registrations include
   `hyremote-release-readiness-metadata`, and that gate matches the declaration with
   `project[ \t\r\n]*\([ \t\r\n]*HyRemote[ \t\r\n]+VERSION[ \t\r\n]+...`
   (`tests/release-readiness/check_release_metadata.cmake:73-78`). Its character classes cover spaces and tabs
   but not newlines, so the previous multi-line declaration is rejected as
   `release-readiness: root project(VERSION ...) is missing or malformed` - which is the P1 in the automated
   review above, and which I reproduced on the pristine tree before changing anything. With the declaration on
   one line the gate passes:
   `-- HyRemote release-readiness metadata gate: PASS (project 0.0.0, ...)`.
2. **The examples block resolves the Qt components it selects on.** `examples/CMakeLists.txt` chooses each
   example with directory-scoped checks such as `if(TARGET Qt6::Widgets)`; those imported targets are created
   inside `src/remoteaccess` and `integrations/qml`, so the sibling scope cannot see them. Verified on this
   tree: without the resolution, a configure with `-DHYREMOTE_BUILD_EXAMPLES=ON` produces **no example target
   at all** (the examples build directory contained only `CMakeFiles/`); with it, `widgets-basic`,
   `quick-basic`, `qml-basic` and `remote-support-showcase` are configured. Without this line the enable flag
   silently means nothing.

## Verification (local, exact Qt 6.8.3 msvc2022_64, clean tree)

```
cmake -S . -B build/p0 -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_PREFIX_PATH=<Qt 6.8.3 msvc2022_64> -DHYREMOTE_BUILD_TESTS=ON
cmake --build build/p0 -- -k 0
ctest --test-dir build/p0 -R "release-readiness|release-profile"
```

- configure: `Configuring done` / `Generating done` (before: `Unknown CMake command
  "hyremote_apply_release_profile"`).
- build: clean, including the Core and RemoteAccess test targets.
- readiness suite: **18/18 pass, 0 fail** - the metadata gate (with the declaration fix), the documentation
  layout gate, the repository layout gate, the package-acquisition, deploy-helper, relocation, consumer and
  release-authority gates, plus all eight milestone release-profile cases (`0.0.1.0` C++-only through
  `1.0.0.0` all modes, negative cases via `WILL_FAIL`).

## Downstream failures I am deliberately NOT including, with their fixes already verified

These are not artifacts of the root rewrite - each reproduces on a tree whose root configure already works -
but they are outside the authorized scope of this repair, so they are offered rather than bundled:

1. `integrations/qpa/hyremote_platform_plugin.cpp`: `<QtGui/private/qplatformintegrationfactory_p.h>` does not
   exist in the exact Qt 6.8.3 kit (the header ships under `QtGui/qpa/`, the form the four neighbouring
   includes already use), and the file overrides `populateLightSystemPalette(QPalette &)` without including
   `QPalette`.
2. `integrations/qpa/tests/CMakeLists.txt`: the two composite tests compile sources that include the
   RemoteAccess private component-factory declarations (`QHostAddress`, generated `RemoteAccessExport.h`) but
   have no include surface for them; linking `HyRemote::RemoteAccess` - as `hyremote-qpa-remote-config-test`
   already does - fixes both.
3. `examples/qml-basic/CMakeLists.txt`: `qt_add_qml_module` needs `CMAKE_AUTOMOC` set before the target
   exists; today a QML+examples configure aborts at `examples/qml-basic/CMakeLists.txt:5`. This is the first
   failure the rerun after this repair will show, so it is worth knowing it is already diagnosed.
4. `hyremote-qpa-platform` still cannot link: it references three Core symbols
   (`CpuFrameStorage::createSinglePlane`, `CpuFrameStorage::mutablePlane`, `hyremote::mapPointerToTarget`).
   Adding `HyRemote::Core` links - and takes the suite to 61/63 with both QPA smoke tests green - but your
   repository-layout gate rejects the link (`QPA platform payload must not link Core directly`), so this is
   the design question from my earlier comment, not a link flag.

Say which of 1-4 you want as bounded PRs against this branch and I will submit them one at a time; otherwise
they stay with the #106 line.

skawu (commented):
## Owner adjudication — #120 and the remaining QPA design question

The latest narrowing is accepted in principle. Keep #120 as the **P0 root-buildability repair**, not as the container for the downstream QPA/QML fixes.

### Disposition of #120

The repaired root behavior is the right recovery shape: restore the pre-`0535f87` configure/install/test-registration semantics, preserve the #111 product description, use the existing `hyremote_validate_release_profile(...)` contract, and do not invent `hyremote_apply_release_profile()`.

Do not add the three previously removed QPA/QML source fixes back into this PR. The root-level Qt component resolution for examples is accepted only because it has direct evidence that `HYREMOTE_BUILD_EXAMPLES=ON` otherwise produces no example targets on this tree; do not expand beyond that.

The one-line `project(...)` declaration is acceptable for this P0 repair as a bounded compatibility measure with the current metadata gate. **Do not turn that formatting workaround into a general source-format rule.** A later gate-hardening change may make the metadata parser whitespace-robust, but do not reopen this P0 or collide with #121's upcoming metadata-path edit just for formatting.

Acceptance for #120 is narrower than “all V1 checks green”: the current head must actually configure, the restored release-readiness/profile registrations must execute, and no regression may be attributable to the root repair. Known downstream QPA/QML failures do not need to be hidden inside this P0. Once the current-head jobs establish that boundary, integrate this repair into #106 and continue from fresh executed failures.

### QPA/Core design ruling — choose a **private RemoteAccess runtime seam**, not a public API and not a direct Core link

The forbidden experiment is confirmed forbidden: `qhyremote` must **not** link `HyRemote::Core` directly. Also reject any solution that adds frame-storage or coordinate-mapping methods to the public `HyRemote::RemoteAccess` application API.

For V1, use the existing pattern already visible in `src/remoteaccess/src/detail/component_factories.hpp`: a **source-private `HyRemote::detail` integration seam implemented/exported by the shared RemoteAccess runtime**. The QPA module may consume that private seam because it is a package-owned peer payload, but the seam must not be installed as a public SDK header or become an application-facing target/API.

Concretely, move/wrap the concrete Core-owned operations needed by QPA behind RemoteAccess-private helpers:

- writable CPU RGBA frame/storage creation needed by composite capture;
- normalized pointer-to-target mapping needed by composite input.

The helpers should return/use only the minimum internal data needed by QPA and keep concrete Core allocation/mapping ownership inside the RemoteAccess binary. `qhyremote` remains linked to `HyRemote::RemoteAccess` + Qt only. Do not solve this by duplicating Core algorithms in QPA, making Core header-only, exporting Core as a second SDK target, or weakening `check_repository_layout.cmake`.

This is **not** a fourth product API. It is an internal ABI between two payloads in the same HyRemote package and may use the existing private export mechanism where Windows DLL visibility requires it.

### Downstream work order after #120 lands

1. **QPA compile plumbing PR:** fix the Qt 6.8.3 `qpa/qplatformintegrationfactory_p.h` include + missing `QPalette`, and fix the two QPA composite test targets to consume `HyRemote::RemoteAccess` like the neighboring test. These are mechanical and may be one bounded PR.
2. **QML example PR:** fix the `qt_add_qml_module`/AUTOMOC ordering independently.
3. **QPA runtime-seam PR:** implement the private RemoteAccess seam above and remove the concrete Core-symbol dependency from the platform module. This PR must prove the repository-layout gate still forbids direct Core linkage and that QPA smoke/native-semantics tests pass through the shared runtime.

Do not mix these three scopes. Fresh hosted failures after the P0 integration remain the execution truth and may reorder the mechanical items if they expose an earlier blocker.

skawu:
## Fresh hosted execution is in - failure attribution, job by job

The ruled order asks for fresh executed CI before the follow-ups; it has now produced results, and I read every
failing job. **None of the failures is attributable to this PR**, which changes only the root `CMakeLists.txt`:

| Job | Result | Cause |
| --- | --- | --- |
| Linux x86_64 / qxcb / QPA module + shared facade | fail | `integrations/qpa/hyremote_platform_plugin.cpp:9:10: fatal error: QtGui/private/qplatformintegrationfactory_p.h: No such file or directory` - **confirms the diagnosed QPA private-header item**, i.e. the authorized QPA plumbing follow-up, not this repair |
| Windows x86_64 / qwindows / QPA module + shared facade | fail | the QPA module **compiles** on that runner (so the private-header path is a real failure only on the Linux install layout), then 13/20 tests fail in the deploy-helper fixtures: `run_deploy_helper_fixture.cmake:112` expects **exactly one** generated deploy script, but the Windows job uses a multi-config generator, so it finds four (`...-Debug/-MinSizeRel/-RelWithDebInfo/-Release.cmake`) |
| Windows x86_64 / Qt 6.8.3 / shared facade | fail | toolchain provisioning: `You have discovered a bug in aqt` then `exit code 254` - an aqtinstall failure, infrastructure |
| Linux x86_64 / fixed shared facade / installed + source | fail | the clean-deployment consumer runs and dies with `libQt6Widgets.so.6: cannot open shared object file` (exit 127) after `build-consumer-installed/cmake_install.cmake:70 (include)` - the deployed tree does not resolve Qt's Widgets runtime |
| Linux x86_64 / Qt QML 6.8.3 / shared RemoteAccess | fail | `43/43 tests passed`, then the example launch fails with `no Qt platform plugin could be initialized` (exit 134) - environment/fixture, the run needs an offscreen or virtual platform |
| Linux/Windows maintained VNC client, Linux/Windows shared facade (native), Linux Quick/Widgets shared RemoteAccess | pass | unchanged paths, no regression from this repair |

Two of these are **new** relative to the items you already delegated and look like genuine defects rather than
infrastructure, so they need your call rather than my assumption:

1. **QPA deploy-helper fixtures assume a single-config generator.** On the Windows multi-config job the fixture
   asserts "exactly one generated QPA/runtime deploy script" and finds four per-configuration ones. Either the
   fixture must select the active configuration or the probe must use a single-config generator; as written it
   can never pass on Windows.
2. **The clean-deployment consumer cannot resolve Qt.** The installed/source fixture is supposed to prove that a
   deployed application runs without SDK/build-tree path assistance - it now fails on `libQt6Widgets.so.6`.
   Either `hyremote_deploy()` does not place the Qt runtime for that fixture, or the fixture's environment is
   wrong. This one touches the V1 clean-deployment claim, so classifying it correctly matters.

What I would like to know:

- integrate **#120** so the authorized queue can start from a head that actually configures (the repair itself
  is verified locally: 18/18 readiness and release-profile tests pass, and the metadata gate's one-line form is
  the P1 workaround you already accepted, with the formatting-invariant gate fix staged in #121);
- whether items 1 and 2 above are delegated to me as bounded PRs, or stay with the #106 line;
- nothing else in this PR is outstanding. The QPA plumbing PR (private-header path plus `QPalette`, and the two
  composite tests' include surface) is ready to go the moment #120 is in.

skawu:
## The whole authorized chain is implemented and locally verified: 64/64 green, QPA links per the frozen rule

While #120 waits, I built the authorized queue locally on one stack (not pushed - your sequencing puts it after
#120) and verified it on the exact Qt 6.8.3 kit with QML + QPA + examples + tests enabled:

- the P0 restore itself (`02c2df1`, which is this PR);
- the QPA plumbing: the `qplatformintegrationfactory_p.h` path that the exact kit actually ships (under
  `QtGui/qpa/`) and the missing `#include <QtGui/QPalette>`, plus the two composite tests' `HyRemote::RemoteAccess`
  include/generated-header surface;
- `examples/qml-basic`: `CMAKE_AUTOMOC` before `qt_add_qml_module`;
- **the QPA/Core seam exactly as ruled**: `src/remoteaccess/src/detail/qpa_composition_seam.{hpp,cpp}` in
  `namespace HyRemote::detail`, source-private, wrapping (1) writable CPU RGBA frame-storage creation and
  (2) normalized pointer mapping. The QPA payload now calls the seam instead of reaching into Core; no Core
  duplication, no Core link, no public `RemoteAccess` API addition, Core stays inside the runtime binary.

Verification results:

- configure and build clean; `plugins/platforms/qhyremote.dll` produced;
- the plugin's link line is `HyRemoteRemoteAccess.lib` + `Qt6Gui/Network/Core` + system libraries - **no Core**;
- `check_repository_layout.cmake` **PASS** ("one shared RemoteAccess runtime across integrations") and
  `check_release_metadata.cmake` **PASS**;
- `ctest` **100% - 64 tests passed, 0 failed**, which includes the QPA smoke tests that could not run before;
- `cmake --install` installs exactly `HyRemote/RemoteAccess.h` and `HyRemote/RemoteAccessExport.h` - the seam is
  **not** part of the SDK surface, and the public C++/QML contract is unchanged.

Nothing in this touches #120 (still one file). The four bounded PRs (QPA plumbing / QML example / QPA-Core seam
/ #121's gate-and-docs migration) open the moment #120 is integrated.

Still open from my previous comment: the two hosted failures that are not explained by the items you delegated -
the Windows multi-config deploy-helper fixture expecting exactly one generated deploy script, and the Linux
installed/source consumer failing to resolve `libQt6Widgets.so.6`. Both need your delegation or a decision.

skawu:
## One decision, then the queue can move

Your sequencing is `#120 -> fresh executed CI -> #122 -> #121 rebase + gate/docs migration -> bounded QPA
plumbing / QML example / QPA-Core seam follow-ups`, and fresh executed CI is now in - I read every failing job
and none of them is attributable to this PR (comment `5714016357`), while the whole authorized chain reaches
`ctest` 64/64 locally with the QPA plugin linked to `HyRemoteRemoteAccess.lib` + Qt only (comment
`5714067198`).

The queue is prepared and pushed, but **no follow-up PR has been opened**, because your ruling conditions them
on #120 being integrated and I do not want to jump the sequence:

- `fix/v1-qpa-plumbing` - the Qt 6.8.3 `qplatformintegrationfactory_p.h` path, the missing `QPalette` include,
  and the two composite tests' `HyRemote::RemoteAccess` include surface (one PR);
- `fix/v1-qml-example-automoc` - `CMAKE_AUTOMOC` before `qt_add_qml_module` (independent);
- `feat/v1-qpa-core-seam` - the ruled source-private `HyRemote::detail` seam wrapping writable CPU RGBA
  frame-storage creation and normalized pointer mapping; contains the plumbing commits because the module
  cannot compile without them, so it should be opened after the plumbing PR.
- `docs/bilingual-user-docs` additionally carries the staged gate-and-docs migration (`f504d98`), which must
  wait for both #120 and #122 because #122 writes `README.md` first.

All three code branches are stacked on this PR's commit, so their CI runs on a configurable tree; once #120 is
integrated, a rebase drops that stacked commit from each diff.

**Requested decision:** may I open those bounded PRs now, stacked on this branch and rebased after the merge, or
do you want them only once #120 is in? Either answer is fine - I am asking because the ordering is yours, not
because I need to proceed. If the answer is "after the merge", the only thing blocking the whole queue is the
integration of #120 itself; PR #120 is one file, `MERGEABLE`, verified locally, and its red checks are all
attributed to pre-existing or downstream causes above.

skawu:
## Owner ruling on fresh hosted evidence and the newly exposed blockers

The fresh hosted execution is valid repository evidence. It confirms the root-CMake P0 repair is doing its job: jobs now progress past the previous `hyremote_apply_release_profile()` configure failure and expose downstream defects. **#120 itself is accepted for integration into the #106 convergence line as the bounded P0 recovery.** The remaining failures do not justify expanding #120 again, and this is not a V1 release-acceptance claim.

### 1. Windows multi-config deploy-helper failure — fixture defect; fold into the authorized QPA plumbing follow-up

The assertion that exactly one generated deploy script must exist is wrong for a multi-config generator. `file(GENERATE ...-$<CONFIG>.cmake)` is allowed to produce one script per configuration. Do **not** force the Windows acceptance job to a single-config generator merely to satisfy the fixture.

Disposition:

- keep the production multi-config behavior;
- make the deploy-helper fixture configuration-aware: on a multi-config generator validate the selected/expected configuration script (and, where useful, assert the generated configuration set is internally consistent); on a single-config generator retain the single-script expectation;
- add deterministic coverage so both generator shapes cannot regress silently;
- if inspection shows a production deploy script is actually wrong for one configuration, fix that product defect explicitly; otherwise this remains test plumbing.

Because this touches the same QPA deploy-helper test surface as the already-authorized QPA plumbing work, include it in that bounded PR rather than opening another competing writer on `integrations/qpa/tests/**`.

### 2. Linux clean deployment cannot resolve `libQt6Widgets.so.6` — V1 deployment-contract defect, separate release-blocking PR

This is **not** an acceptable environment workaround. The frozen deployment contract says one `hyremote_deploy(TARGET ...)` call owns the normal deployed application shape: Qt's deployment machinery places the Qt runtime, while the HyRemote supplemental step places the shared facade/integration payloads. The clean consumer is deliberately executed with SDK/build-tree path assistance removed.

Therefore:

- keep the environment-scrubbed runtime test;
- do **not** add `LD_LIBRARY_PATH`, point the test back at the Qt SDK, or weaken the clean-deployment claim;
- diagnose why Qt deployment does not carry the transitive Qt runtime required by `HyRemoteRemoteAccess` on Linux, and repair `hyremote_deploy()` / its Qt deployment integration so the installed tree is self-contained under the documented contract;
- validate installed and source acquisition at minimum; extend to QML/QPA consumers wherever the same helper path is shared;
- preserve the artifact rule: one shared `HyRemoteRemoteAccess`, internal Core, no second SDK personality.

Delegate this as a **separate bounded V1 deployment PR** after #120 lands. It is a #104 blocker until the clean deployed consumer runs with the SDK/runtime search environment scrubbed.

### 3. Linux QML example platform-plugin failure — hosted fixture/environment, not a product semantic change

The run that reports `43/43 tests passed` and then fails because no Qt platform plugin can initialize has already proved the library/test portion. The example launch still needs a valid headless GUI environment. Use the repository's existing virtual-display/offscreen convention (`xvfb-run` for GUI E2E; `QT_QPA_PLATFORM=offscreen` only where the scenario is genuinely non-visual). Do not change HyRemote product behavior to accommodate a headless runner.

This may travel with the already-authorized QML-example reliability PR because it is the same E3 example acceptance surface, provided the diff remains limited to the example/build/hosted fixture and does not alter product semantics.

### 4. aqtinstall failure — infrastructure

`aqt` exiting with its own internal error is infrastructure/tool provisioning evidence. No HyRemote code change is authorized from that failure. Re-run only on a stable new head; do not retry-spam or relax product gates.

### Sequence / ownership

1. integrate #120 into `feature/104-v1-ga-acceptance-matrix`;
2. rerun/read back the repaired-head evidence;
3. #122 may integrate next on its docs surface, then #121 rebases as already ruled;
4. code follow-ups may proceed on non-overlapping surfaces: QPA plumbing + multi-config fixture, QML example reliability, QPA private runtime seam, and the separate clean-deployment runtime fix;
5. one writer per surface remains mandatory.

The local 64/64 stack is useful diagnosis and confidence, but hosted/reference execution on the integrated head remains the evidence authority for #104.

PR #124, opened 2026-09-17
build(qpa): compile against the qualified Qt headers and make the deploy fixture configuration-aware

skawu:
## What this fixes

The two QPA defects the fresh hosted execution exposed on the repaired head, plus the multi-config deploy
fixture defect you folded into this surface. Scope is `integrations/qpa/**` only; no product behaviour, no
public API, no CMake option and no compatibility row changes.

1. **Qt 6.8.3 private-header path.** `hyremote_platform_plugin.cpp:9` included the integration factory as
   `QtGui/private/qplatformintegrationfactory_p.h`; the exact qualified kit ships it in the QPA header set
   (`QtGui/qpa/`), which is also the form the four neighbouring includes in the same file already use. The
   `private/` path does not exist there, so the module failed to compile on every platform - this is the
   Linux failure from the fresh run (`qplatformintegrationfactory_p.h: No such file or directory`).
2. **Missing `QPalette` include** in the same translation unit (`populateLightSystemPalette(QPalette &)`).
3. **Composite test include surface.** `hyremote-qpa-composite-capture-test` and
   `hyremote-qpa-composite-input-test` compile sources that include the RemoteAccess private
   component-factory declarations, which need `QHostAddress` and the generated `HyRemote/RemoteAccessExport.h`.
   Both now link `HyRemote::RemoteAccess`, exactly as the neighbouring `hyremote-qpa-remote-config-test`
   already does.
4. **Deploy-helper fixture, multi-config defect (your item 1).** The fixture asserted that exactly one
   generated deploy script must exist, which is wrong for a multi-config generator: `hyremote_deploy()` names
   its script `...-$<CONFIG>.cmake` and `file(GENERATE)` expands that once per configuration. **Inspected
   production behaviour first**: the three scripts a multi-config generator produces differ only in each
   configuration's own payload locations (`.../<Config>/qpa-probe.dll`), which is correct - so this is test
   plumbing, not a product defect. The runner now:

   - detects the generator shape from the fixture's own `CMakeCache.txt`;
   - single-config: keeps the single-script expectation and additionally cross-checks the script against the
     configured `CMAKE_BUILD_TYPE`;
   - multi-config: expects exactly one script **per declared configuration**, asserts the generated set is
     internally consistent (every configuration's script either references its own configuration location or
     none does, so no configuration can deploy another's payload), checks the universal fragments
     (`qt_deploy_runtime_dependencies`, `ADDITIONAL_LIBRARIES`) in **every** configuration's script, and runs
     the full assertions on a selectable configuration;
   - two new registrations pin both shapes explicitly (`Ninja` and `Ninja Multi-Config`) so neither can
     regress silently on any host with ninja; the pinned generator gets `CMAKE_MAKE_PROGRAM` forwarded so the
     coverage does not depend on the test process PATH.
   Per your instruction the Windows acceptance job was **not** moved to a single-config generator.

## Evidence (exact Qt 6.8.3 kit, `-DHYREMOTE_BUILD_TESTS=ON -DHYREMOTE_BUILD_QML_API=ON -DHYREMOTE_WITH_QPA_PROXY=ON`)

- configure succeeds (it failed at `hyremote_apply_release_profile()` before #120, and this module failed to
  compile before this PR);
- every target builds except the `hyremote-qpa-platform` link, which is exactly what the QPA/Core seam PR
  completes - that link is the ruled reason `HyRemote::Core` must not be added here;
- **18/18 deploy-helper tests pass**, including the new single-config and multi-config registrations, verified
  with ninja removed from the test PATH;
- negative controls hold: an out-of-set expected configuration is rejected, and the multi-config set
  assertions exercise all three declared configurations.


skawu (commented):
Owner review of the actual patch, not only the PR description.

**Disposition: QPA plumbing shape accepted with one required factual correction before integration.**

Accepted scope:
- qualified Qt 6.8.3 QPA header include + explicit `QPalette` include;
- the two composite tests receiving the `HyRemote::RemoteAccess` include/generated-header surface;
- generator-aware deploy-helper fixture that preserves the real multi-config production shape instead of changing Windows acceptance to a single-config generator;
- explicit single-config + multi-config fixture coverage.

Required correction: the new comment in `hyremote_platform_plugin.cpp` says the `QtGui/private/...` include caused the QPA module to fail to compile **"on every platform"**. Fresh hosted evidence did not establish that: Linux failed at that include, while the earlier Windows run got past QPA compilation and failed later in the multi-config deploy fixture. Keep the portable/qualified fact, but remove the universal claim. State simply that the exact qualified Linux/Qt layout exposes the factory through the QPA header set and that the `qpa/` form is the qualified cross-platform source form used by this module.

Integration evidence requirement is bounded: the current QPA run must execute far enough to prove the old header-path failure and the multi-config fixture failure are gone. A later link failure caused by the already-known QPA/Core boundary is not attributable to this plumbing PR and must not be 'fixed' here by linking Core.

After that correction/evidence, #124 is the first code follow-up to integrate. Do not absorb the private Runtime seam or clean-deployment Qt-runtime repair into this PR.

skawu:
Absorbed into the single V1 convergence line PR #106. I verified that the four QPA files had no competing changes on #106 since this PR's base, then applied their exact blobs atomically on top of the current candidate as commit `d93b6ff6204151a26fdde248e2119b5103defa7e`. This brings in the exact-Qt 6.8.3 QPA header fix, QPalette include, composite-test RemoteAccess linkage and configuration-aware deploy-helper fixture without merging a parallel branch or linking QPA to Core. Closing this PR as superseded by the canonical convergence line.

PR #125, opened 2026-09-17
build(examples): enable AUTOMOC before qt_add_qml_module

skawu:
## What this fixes

`examples/qml-basic` cannot be configured with examples enabled:
`qt_add_qml_module()` extracts QObject metadata while configuring the target, so `CMAKE_AUTOMOC` has to be
on before the target exists.

Control experiment (the AUTOMOC line removed, everything else identical):

```
CMake Error at C:/Qt/6.8.3/msvc2022_64/lib/cmake/Qt6Qml/Qt6QmlMacros.cmake:3689 (qt6_extract_metatypes)
Call Stack: ... qt6_add_qml_module -> examples/qml-basic/CMakeLists.txt:5 (qt_add_qml_module)
-- Configuring incomplete, errors occurred!
```

Reproduced on the integrated head `377c608` with `-DHYREMOTE_BUILD_EXAMPLES=ON`. This matters for the
`V1 GA acceptance` job, which is the one hosted job that enables examples, so the E3 example acceptance
surface currently cannot configure there.

The target now sets `CMAKE_AUTOMOC` before `qt_add_qt_executable`, the same rule
`integrations/qml/HyRemote` already follows for the module itself. Scope: one example's CMakeLists; no
product code, no API, no option, no compatibility row.

Your ruling also allowed the headless-launch part of item 3 (no Qt platform plugin can initialize on the
runner) to travel with this PR. **It does not**: that fix belongs to the hosted fixture/example launch
environment (`xvfb-run` / offscreen as your ruling specifies), and I would rather keep this diff to the one
configure defect and let you place the environment change deliberately, instead of mixing a workflow edit
into a build fix.

## Evidence

- Exact Qt 6.8.3 kit, `-DHYREMOTE_BUILD_EXAMPLES=ON -DHYREMOTE_BUILD_QML_API=ON`: the example configures and
  the target builds (before this change the whole configure aborted at that file);
- the full stack containing this commit builds and reaches `ctest` 66/66;
- independent of the QPA plumbing / seam PRs (no shared file).


skawu (commented):
Owner review found a scope/ancestry problem in the **actual PR diff**.

**Disposition: HOLD in the current shape; rebase/recreate before integration.**

The intended change is correct: `examples/qml-basic/CMakeLists.txt` must enable `CMAKE_AUTOMOC` before `qt_add_qml_module`/target creation, and that bounded one-file fix is authorized.

However, this PR currently also presents the entire #120 root `CMakeLists.txt` recovery as part of its diff. That contradicts the PR description and the one-writer/bounded-slice rule. #120 is already integrated into #106; it must not be re-carried through the QML example PR.

Required action:
- rebase or recreate the branch on the current `feature/104-v1-ga-acceptance-matrix` head;
- verify the PR diff reduces to **only `examples/qml-basic/CMakeLists.txt`** (unless fresh evidence proves an immediately coupled file is strictly required);
- keep the headless Linux launch/environment correction out of this PR; that is a separate workflow/fixture concern.

The current Transparent QPA failure is not evidence against the AUTOMOC change: the executed Linux job still fails at the old `QtGui/private/qplatformintegrationfactory_p.h` baseline because #124 is not in this branch. Do not widen #125 to solve that unrelated prerequisite.

Once the diff is clean and the QML/example configure/build evidence executes on the rebased head, this one-file fix is acceptable independently of #126.

skawu:
Absorbed into the single V1 convergence line PR #106. The real configure blocker is valid: `examples/qml-basic` must enable AUTOMOC before `qt_add_qml_module()`. The canonical branch now carries that fix as commit `d988b690086a81ca0c1ef6e229e5767db0b30c29`. Keeping this separate PR open would recreate a parallel V1 implementation branch, so this PR is superseded and should not merge independently. The hosted launch-environment issue remains owned by #106/#104, not by this one-file build fix.

PR #126, opened 2026-09-17
refactor(qpa): route composite capture and input through a source-private RemoteAccess seam

skawu:
## What this implements

The ruled QPA/Core boundary repair: `qhyremote` must not link `HyRemote::Core`, the repository-layout gate
must not be weakened, and no frame-storage or input-mapping method may be added to the public
`HyRemote::RemoteAccess` API. The seam is therefore **source-private** and lives in the shared runtime:

- `src/remoteaccess/src/detail/qpa_composition_seam.{hpp,cpp}` in `namespace HyRemote::detail`, declared with
  `HYREMOTE_REMOTEACCESS_EXPORT` and compiled into the shared runtime. It wraps exactly the two Core-owned
  operations the composite path needs: writable CPU RGBA frame-storage creation plus writable first-plane
  access, and normalized pointer mapping for composite input;
- `integrations/qpa/composite_target.cpp` and `integrations/qpa/interactive_composite_target.cpp` call the
  seam instead of reaching into Core. Core ownership stays inside the RemoteAccess binary; no Core
  duplication, no header-only Core, no second SDK personality.

`integrations/qpa/hyremote_platform_plugin.cpp` keeps its own local change under the QPA plumbing surface
(different file), so this PR does not produce a second writer there.

## Evidence (exact Qt 6.8.3 kit, tests + QML API + QPA enabled, examples off)

- configure and build clean; `plugins/platforms/qhyremote.dll` produced;
- the plugin's link line is `HyRemoteRemoteAccess.lib` + Qt + system libraries - **no Core**;
- `check_repository_layout.cmake` **PASS** ("one shared RemoteAccess runtime across integrations") and
  `check_release_metadata.cmake` **PASS**;
- **`ctest` 66/66 pass**, including the QPA smoke tests and the composite capture/input tests that could not
  run before this change;
- `cmake --install` still installs only `HyRemote/RemoteAccess.h` + `HyRemote/RemoteAccessExport.h`, i.e. the
  seam header is **not** part of the installed SDK surface and the public C++/QML contract is unchanged.

## Dependency

This branch contains the QPA plumbing commits it needs (the module cannot compile or link without them), so
it shows a diff that overlaps that PR. **Merge/open the QPA plumbing PR first**; after it is integrated a
rebase reduces this diff to the seam alone.


skawu (commented):
Owner architecture review of the actual seam patch.

**Disposition: architecture accepted in principle; integration waits on #124 and a seam-only rebase.**

The implementation follows the frozen ruling:
- Core-owned `CpuFrameStorage` allocation/writable access and pointer mapping are implemented inside the shared RemoteAccess runtime;
- QPA calls a source-private `HyRemote::detail` seam;
- no public `HyRemote::RemoteAccess` API is added;
- the seam header is source-private and not installed;
- no Core algorithm is duplicated into QPA and no second SDK target/personality is created.

The current PR intentionally contains #124's plumbing commits. That is acceptable as a temporary stacked development branch, but it is **not** the integration shape. Integrate #124 first, then rebase #126 onto the updated #106 head so this PR shows only the seam/runtime-composition delta.

On that rebased head, required acceptance evidence is:
1. `qhyremote` links to `HyRemote::RemoteAccess` + Qt/system libraries and **not** `HyRemote::Core`;
2. repository-layout gate remains green without weakening its direct-Core prohibition;
3. QPA smoke/native-semantics and composite capture/input tests pass;
4. installed SDK still does not contain `qpa_composition_seam.hpp` or export Core/QpaPlatform as application targets;
5. Windows DLL symbol visibility works through the private exported seam.

The reported local 66/66 result is good diagnostic evidence, and the RemoteAccess hosted workflow on the current stacked head has already succeeded, but the Transparent QPA workflow is still executing. Do not describe this PR as fully hosted-green until the rebased seam-only head executes.

No architecture redesign is requested; only dependency-order/rebase hygiene and executed evidence remain.

skawu:
Absorbed into the single V1 convergence line PR #106. After #124's QPA plumbing was integrated, I compared `4a374a3c... -> 9f81a530...` and isolated exactly five seam-only files. Those were applied atomically on top of the current candidate as commit `0de0fb2b709543764b9a7e39dcc6dc36a6ba9f95`. The result preserves the ruled boundary: Core remains static/private behind the one shared `HyRemote::RemoteAccess` runtime; the QPA MODULE calls a source-private exported seam; the seam header is not installed; no public `RemoteAccess` API was enlarged and QPA does not link `HyRemote::Core`. Closing this parallel PR as superseded by the canonical convergence line.

PR #127, opened 2026-09-17
fix(deploy): place the payload's private Qt runtime in deployed applications

skawu:
## What this repairs

Your item 2: the Linux clean deployment could not resolve `libQt6Widgets.so.6`. The failure, verbatim from
the hosted job (`SDK consumption` -> `Linux x86_64 / fixed shared facade / installed + source`, step
`env -u LD_LIBRARY_PATH deploy-installed/bin/hyremote-installed-consumer`):

```
deploy-installed/bin/hyremote-installed-consumer: error while loading shared libraries: libQt6Widgets.so.6:
cannot open shared object file: No such file or directory
##[error]Process completed with exit code 127
```

not on a tree where the facade was missing - the same log shows
`-- Installing: .../deploy-installed/lib/libHyRemoteRemoteAccess.so` and the app got far enough to start
loading it.

### Root cause

The consumer finds and links only `Qt6 COMPONENTS Core Network` and `HyRemote::RemoteAccess`; Widgets/Quick are
**private** links of the shared payload, so they are deliberately absent from the consumer's link interface and
from the app's own link line. Qt's deployment machinery is what has to place the Qt runtime, and it scans with
`file(GET_RUNTIME_DEPENDENCIES)`:

1. Qt's generated deploy script scans the **build-tree** application, whose build RPATH still contains the Qt
   installation, so it resolves and installs Core, Network, ICU, translations **and the facade**.
2. Recursing into the facade, `libQt6Widgets.so.6` cannot be resolved. The SDK installs the facade with a
   deliberately narrow runtime path - the same log says
   `-- Set non-toolchain portion of runtime path of ".../libHyRemoteRemoteAccess.so" to "$ORIGIN"` - so the
   facade's own RPATH is the SDK/deploy lib directory and never points at the Qt installation that built the
   SDK. The dependency lands in `UNRESOLVED_DEPENDENCIES` and is never placed.
3. The HyRemote supplemental step has the same blind spot: it resolves its `ADDITIONAL_LIBRARIES` entry
   against the deployment prefix, i.e. the just-installed facade, whose runtime path is `$ORIGIN` for the same
   reason.

`windeployqt` resolves through the Qt installation instead of through the binary's runtime path, which is why
only the Linux generic deploy tool was affected - the same shape deployed and ran on Windows.

### The repair

- The SDK records the adapter set the shared payload was built with **inside the installed deploy helper**
  (`HYREMOTE_DEPLOY_PAYLOAD_QT_MODULES`, configured from the same adapter flags that already drive the
  payload's sources), and the source-tree helper answers the same question from the build's own adapter flags.
- `hyremote_deploy()` places those Qt modules beside the facade and passes them to
  `qt_deploy_runtime_dependencies`, so those modules' own Qt dependencies are deployed as well. Both the plain
  runtime path and the QPA path share this helper, so both are covered.
- On Unix the module payload is installed with `FOLLOW_SYMLINK_CHAIN`, matching Qt's own deployment, so the
  Soname the facade requests and the file the loader opens are both present.

**The list lives in the installed helper, not in a new package variable**, so the frozen installed package
surface is unchanged (`hyremote_deploy()` stays the single entry point and no new consumer-visible variable
appears) - the `hyremote-release-readiness-deploy-helper-contract` guard still passes.

Nothing in the diagnosis path is weakened, exactly as ruled: no `LD_LIBRARY_PATH`, no pointing the test back at
the Qt SDK, no change to the environment-scrubbed runtime test, one shared `HyRemoteRemoteAccess`, internal
Core, no second SDK personality.

### Evidence

- installed-SDK consumer flow reproduced locally on the exact Qt 6.8.3 kit: the SDK install now carries
  `set(HYREMOTE_DEPLOY_PAYLOAD_QT_MODULES "Widgets;Quick")`, and the generated deployment script contains

  ```
  file(INSTALL DESTINATION "${QT_DEPLOY_PREFIX}/${QT_DEPLOY_BIN_DIR}" TYPE FILE FILES ".../bin/HyRemoteRemoteAccess.dll")
  file(INSTALL DESTINATION "${QT_DEPLOY_PREFIX}/${QT_DEPLOY_BIN_DIR}" TYPE FILE FILES "C:/Qt/6.8.3/msvc2022_64/bin/Qt6Widgets.dll")
  file(INSTALL DESTINATION "${QT_DEPLOY_PREFIX}/${QT_DEPLOY_BIN_DIR}" TYPE FILE FILES "C:/Qt/6.8.3/msvc2022_64/bin/Qt6Quick.dll")
  qt_deploy_runtime_dependencies(
      EXECUTABLE "${QT_DEPLOY_BIN_DIR}/hyremote-installed-consumer.exe"
      ADDITIONAL_LIBRARIES "${QT_DEPLOY_BIN_DIR}/HyRemoteRemoteAccess.dll" "C:/Qt/6.8.3/msvc2022_64/bin/Qt6Widgets.dll" "C:/Qt/6.8.3/msvc2022_64/bin/Qt6Quick.dll"
  )
  ```

- installing that consumer deploys `Qt6Widgets.dll` and `Qt6Quick.dll` into the deployment now (the deployment
  before the change contained only what the application itself links);
- `deploy-helper` fixture suite **16/16 pass**;
- the five deployment-related release-readiness gates pass, including
  `hyremote-release-readiness-deploy-helper-contract` and `-deployment-relocation`;
- the acceptance surface remains hosted: the Linux `SDK consumption` clean-deployment step (`env -u
  LD_LIBRARY_PATH ...`) is what will confirm the fix on the integrated head.

### One deliberate omission

I did **not** add behavioural fixture coverage for the payload-runtime contract, because that fixture lives on
`integrations/qpa/tests/**`, which is the surface of the open QPA plumbing PR (#124) - one writer per surface.
I have the fixture change prepared (a knob that advertises a payload Qt module, asserts the generated script
installs and hands over that module, plus a negative case proving that a consumer Qt which cannot supply it
fails closed with a clear message) and will fold it in on your word once #124 is integrated. If you prefer it
in this PR instead, say so and I will move it here.


skawu:
Superseded by `e1a67ac` on the convergence line, which I have reviewed and which takes a strictly more general
approach than this PR: it discovers the facade's private runtime dependencies through
`file(GET_RUNTIME_DEPENDENCIES ... DIRECTORIES <consumer Qt prefix>)` instead of needing the SDK to record which
adapters it was built with, and it copies only dependencies that live under the consumer's Qt prefix. That is a
better contract - no metadata to keep in sync, and it stays correct if the adapter set changes - so this parallel
implementation should not merge. Closing it, not merging.

One blocker in `e1a67ac` is recorded on #106 (comment `5715083063`): the Linux bootstrap emits
`$<TARGET_FILE_DIR:Qt6::Core>` unconditionally, and the deploy-helper fixture has no Qt targets, so every
`hyremote-qpa-deploy-helper-*` sub-configure fails in the generate step on Linux (`No target "Qt6::Core"`,
reproduced locally). The fix is a configure-time `if(TARGET Qt6::Core)` guard around the emission; I can prepare
that as a one-line bounded patch, or hand it to #106, whichever you prefer.

Kept for the record: this branch contains a working, locally verified alternative (`hyremote_deploy()` installs
the payload's private Qt modules beside the facade and hands them to `qt_deploy_runtime_dependencies`, with the
adapter set baked into the installed deploy helper rather than published as a new package variable) plus the
deferred fixture coverage for that contract. Nothing from it is lost; reopen or ask for a port if you want that
shape instead.

PR #128, opened 2026-09-17
fix(deploy): guard the Linux private-runtime bootstrap on the consumer Qt target

skawu:
## One-line blocker in `e1a67ac`, with the minimal guard

`e1a67ac` is the right repair (it discovers the facade's real private `DT_NEEDED` entries from the consumer's Qt
prefix instead of needing the SDK to publish an adapter list - strictly more general than the approach I had in
#127, which is why I closed that PR as superseded). But on Linux it emits a **target-dependent generator
expression unconditionally**:

```cmake
file(GET_RUNTIME_DEPENDENCIES
    LIBRARIES "${QT_DEPLOY_PREFIX}/${runtime_deploy_dir}/$<TARGET_FILE_NAME:HyRemote::RemoteAccess>"
    DIRECTORIES "$<TARGET_FILE_DIR:Qt6::Core>"
    ...)
```

`file(GENERATE)` evaluates that expression in the project that deploys, and
`integrations/qpa/tests/deploy_helper_fixture` deliberately has **no Qt targets at all** - it stubs
`qt_deploy_runtime_dependencies` and the whole Qt deploy surface. So on Linux every deploy-helper fixture
sub-configure fails in the generate step, before any assertion runs. Minimal reproduction with exactly the
emitted content in a project without Qt:

```
-- Generating done (0.0s)
  No target "Qt6::Core"
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

The same content without the generator expression configures cleanly, so it is the target-dependent expression
rather than the dependency call.

Consequence: the `hyremote-qpa-deploy-helper-*` registrations (registered whenever `HYREMOTE_BUILD_TESTS` is on
inside `integrations/qpa/CMakeLists.txt`, with no platform guard) fail at configure time on **every Linux job
that enables QPA + tests** - `Transparent QPA Proxy`, `V1 GA acceptance`, and any successor. The nine runs on
`e1a67ac` were still `queued` when this was written, so this pre-empts them rather than reporting from them.

## The guard

Emit the bootstrap only when the deploying project actually has `Qt6::Core`:

```cmake
    if(NOT TARGET Qt6::Core)
        set(${output_var} "" PARENT_SCOPE)
        return()
    endif()
```

Every real consumer satisfies it - it cannot link `HyRemote::RemoteAccess` without Qt Core - while fixtures and
any project that only fakes Qt deployment stay valid. I preferred this over giving the fixture a `Qt6::Core`
stand-in, because that would add Qt to a surface that deliberately avoids it (and it is #124's surface).

## Evidence

- local: `deploy-helper` fixtures **16/16 pass** and the five deployment release-readiness gates pass
  (`deploy-helper-contract`, `deployment-relocation`, `consumer-simplicity`, `repository-layout`, `metadata`);
  every fixture includes this helper, so the file itself is exercised by them;
- local: the failure mode above reproduced with the exact emitted content, and the guarded content emits no
  generator expression at all, so the generate step cannot fail for a project without Qt;
- honoured boundary: no new package surface, no second SDK personality, no `LD_LIBRARY_PATH`, no change to the
  environment-scrubbed runtime test;
- acceptance stays hosted: the Linux QPA/GA jobs exercise the generate path and the clean-deployment step.

## Observation, deliberately not changed here

The bootstrap `FATAL_ERROR`s on **any** unresolved dependency of the facade, including non-Qt ones. Qt's own
scan merely reports `UNRESOLVED_DEPENDENCIES`. On the reference runners everything resolves, and failing closed
is defensible for a deployment contract, so I left it alone - but if a future runner misses a system library the
deployment will fail rather than warn, and narrowing the failure to unresolved entries that should have come
from the Qt prefix would keep the strong signal without that exposure. Say the word and I will add it as its own
bounded change.

This branch is one file, 11 added lines, based on `e1a67ac`; absorb it as an exact blob if that is easier than
merging.


skawu (commented):
Owner ruling: **ACCEPT the #128 guard as the correct bounded follow-up to `e1a67ac` and absorb/merge it into #106 now.**

I checked the actual one-file diff against the bootstrap added by `e1a67ac`.

The guard is semantically correct:

- the Linux bootstrap emits target-dependent generator expressions (`HyRemote::RemoteAccess` / `Qt6::Core`) that are evaluated in the deploying project;
- a real installed/source consumer that links the facade necessarily has a real Qt Core target, so the guard does not remove the production bootstrap from a valid consumer;
- the deploy-helper fixture deliberately stubs Qt deployment and carries no real Qt targets, so evaluating `$<TARGET_FILE_DIR:Qt6::Core>` there is a fixture-shape failure, not product deployment behavior;
- returning an empty bootstrap in that no-Qt fixture case preserves the existing mocked deploy-helper contract instead of contaminating the fixture with a fake `Qt6::Core` target.

No API/package/architecture/support claim changes. Keep this PR exactly one-file/bounded; do not add the unresolved-dependency-policy observation here.

**Do not wait for the currently queued `e1a67ac` jobs before absorbing this guard.** The old head now has a deterministic Linux generate-stage defect in the fixture path; those queued runs are stale evidence for the next candidate. After the guard is on #106, let the new-head hosted Linux QPA/GA/SDK-consumption runs establish the authoritative result.

On the observation that the bootstrap currently `FATAL_ERROR`s for any unresolved facade dependency: **no further change is authorized now.** Failing closed is acceptable for the V1 clean-deployment contract on the reference environments. Re-open that policy only if fresh reference-host evidence shows a legitimate system/non-Qt dependency being reported unresolved and causing a false deployment failure. Do not pre-weaken the signal.

Acceptance after absorption remains the existing contract: Linux deploy-helper fixtures generate and execute; clean installed/source C++ deployment resolves the facade's private Qt runtime with SDK/runtime paths scrubbed; QML/QPA combined shapes remain correct; no `LD_LIBRARY_PATH` workaround.

PR #129, opened 2026-09-17
fix(qpa): register the source payload relocation test with the variables it declares

skawu:
## What this fixes

On the current candidate `6e861e4` the QPA Linux job and the V1 GA Linux job each report **three** failures.
Two of them abort with

```
qt.qpa.plugin: Could not load the Qt platform plugin "hyremote" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized.
```

(`hyremote-qpa-proxy-smoke`, `hyremote-qpa-native-semantics`), and the third -

```
CMake Error at integrations/qpa/tests/check_source_payload_relocation.cmake:4 (message):
  QPA_FILE, REMOTEACCESS_FILE and TEST_ROOT are required
```

- is the one that should be telling us *why* the plugin cannot load: that script copies the built payloads
into a scratch tree, rewrites the plugin's RUNPATH, and runs `ldd` under a scrubbed environment, failing on any
`not found`. **It never got that far**, because its registration passes different variable names than the script
declares:

```cmake
-DQPA_SOURCE_FILE=$<TARGET_FILE:hyremote-qpa-platform>
-DREMOTEACCESS_SOURCE_FILE=$<TARGET_FILE:hyremote-remoteaccess>
# and no TEST_ROOT at all
```

Both sides have been in that state since `d9caddd`, so this Linux-only gate has been failing on its own argument
check rather than testing the relocation contract - which also means the V1 GA Linux job cannot pass while it is
broken.

## The fix

Register the test with the variables the script declares, plus a scratch root under the test binary directory:

```cmake
-DQPA_FILE=$<TARGET_FILE:hyremote-qpa-platform>
-DREMOTEACCESS_FILE=$<TARGET_FILE:hyremote-remoteaccess>
-DTEST_ROOT=${CMAKE_CURRENT_BINARY_DIR}/source-payload-relocation
```

## Evidence

- local: running the script with the corrected wiring proceeds **past** the argument check to its path
  assertions (it reports the payload path it was given), i.e. the wiring contract now matches; with the old
  names it stops at line 4, exactly as the hosted log shows;
- local: configure stays clean and `release-readiness-deployment-relocation`,
  `release-readiness-deploy-helper-contract` and the deploy-helper fixtures pass **19/19**;
- the Linux QPA/GA jobs are the acceptance surface for this test.

## What this does *not* claim

I am not claiming the root cause of the two plugin-load aborts. With this registration repaired, that script's
`ldd` step becomes the diagnostic that names the unresolved dependency - which is the point of fixing it first.
If the loader failure is a real product/linkage issue rather than a runner environment gap, that run will say so
explicitly instead of leaving Qt's generic message.

One file, 7 added lines, based on `6e861e4`; absorb it as an exact blob if that is easier than merging.


skawu (commented):
Owner ruling: ACCEPT this bounded test-registration repair and absorb it into #106 before any further Linux QPA loader diagnosis.

The current hosted Linux QPA run on exact candidate `6e861e4` proves the defect directly: `hyremote-qpa-source-payload-relocation` exits at line 4 because the registration does not provide `QPA_FILE`, `REMOTEACCESS_FILE`, and `TEST_ROOT`. I also re-read `check_source_payload_relocation.cmake`; those are exactly the declared inputs and the script then copies/relocates the source payload and runs `ldd`. The submitted diff changes only the registration to those exact names and adds the missing deterministic test root.

Scope ruling:
- keep #129 one-file / registration-only;
- do not fold either `HyRemote QPA Proxy could not create native delegate "xcb"` failure into this PR;
- do not change the relocation script semantics, RPATH contract, or deployment helper here;
- no need to wait for a full matrix to decide whether the current registration is wrong: the authoritative hosted base already executed and demonstrated the argument mismatch. After absorption, the next Linux QPA run must show the relocation test actually reaches and passes/fails its `RPATH_CHANGE`/`ldd` assertions.

This is diagnostic-gate repair, not acceptance of the remaining QPA runtime behavior.

PR #130, opened 2026-09-17
fix(qpa): compare the relocated RemoteAccess path after normalizing it

skawu:
## Ruling item 2, first step: the relocation readback, and what it actually shows

`#129` is integrated, so the repaired gate now runs. On the current candidate its output is:

```
-- Set non-toolchain portion of runtime path of ".../source-payload-relocation/plugins/platforms/libqhyremote.so" to "$ORIGIN/../../lib"
CMake Warning (dev) at integrations/qpa/tests/check_source_payload_relocation.cmake:62 (file):
  Given path: ... does not refer to an existing path on disk.
CMake Error at integrations/qpa/tests/check_source_payload_relocation.cmake:64 (message):
  relocated qhyremote escaped deployment tree: expected
  .../source-payload-relocation/lib/libHyRemoteRemoteAccess.so, got ...
  libHyRemoteRemoteAccess.so => .../source-payload-relocation/plugins/platforms/../../lib/libHyRemoteRemoteAccess.so
```

Two things follow, and they split the diagnostic the way you asked:

1. **Relocation itself holds.** The loader resolves the facade through the anchor the plugin's RUNPATH
   contains (`plugins/platforms/../../lib/...`), and that anchor points inside the deployed tree. There is no
   `not found`, and the gate's own `ldd` "unresolved dependency" check passed before reaching this assertion.
2. **The assertion is the defect.** It compares that anchor-form string with a resolved expectation and tries
   to normalize it with `file(REAL_PATH)`, which keeps the input unchanged - CMake only warns - when the path
   does not resolve as written. A spelling difference therefore became a deterministic failure.

The same run also shows the two `xcb` delegate aborts are **gone** on this head: `hyremote-qpa-proxy-smoke`
and `hyremote-qpa-native-semantics` both pass, so item 2's "unresolved ELF/runtime dependency vs native-plugin
discovery" question is answered by execution - the native delegate loads once the Qt 6.8 xcb runtime
dependencies are pinned, and no product-mode change was needed.

## The fix

Normalize the reported path textually and assert existence separately:

```cmake
string(STRIP "${CMAKE_MATCH_1}" reported_remoteaccess)
cmake_path(NORMAL_PATH reported_remoteaccess OUTPUT_VARIABLE resolved_remoteaccess)
if(NOT EXISTS "${resolved_remoteaccess}")
    message(FATAL_ERROR "relocated qhyremote reported a RemoteAccess path that does not exist: [${reported_remoteaccess}] ...")
endif()
if(NOT resolved_remoteaccess STREQUAL expected_remoteaccess)
    message(FATAL_ERROR "... expected [${expected_remoteaccess}], reported [${reported_remoteaccess}], normalized [${resolved_remoteaccess}] ...")
endif()
```

That keeps the assertion about *which* file the loader selected - the contract under test - instead of about
how the loader spelled the path, and the delimited reported/normalized values make a truncated capture
obvious in the log instead of hiding it behind `REAL_PATH`'s warning.

## Evidence

- local: `cmake_path(NORMAL_PATH ".../plugins/platforms/../../lib/libHyRemoteRemoteAccess.so")` collapses to
  `.../lib/libHyRemoteRemoteAccess.so`, i.e. exactly the expectation; a probe of the previous code path also
  confirms `file(REAL_PATH)` does not normalize a path it cannot resolve (it warns and keeps the input);
- local: the script parses with the registered inputs and reaches its payload checks (`ldd` itself is
  Linux-only, so the last stage is hosted);
- constraints respected: registration-only change before this, no platform substitution, no skip,
  `xcb` stays the Linux reference delegate, and no product/lifecycle semantics touched.

One file, +14/-2 lines, based on `6169528`; absorb it as an exact blob if that is easier.


skawu:
Superseded by `d6e212c`, and I want to be precise about why that one is the right fix and this one is not.

`d6e212c` names the real cause: a POSIX-style character class containing `\n` is unsafe in CMake's regex engine,
where that escape can be read as the literal `n`. That is exactly the truncation visible in the hosted log -
`/home/ru` instead of `/home/runner/...` - so the captured value was never a path at all.

This PR attacked the same failure from the wrong end. It assumed the capture was complete and that
`file(REAL_PATH)` was failing to normalize the `$ORIGIN/../../lib` anchor form. With the truncation present,
normalizing could not have produced the expected path either: my version would only have failed with a clearer
message, not passed. The `EXISTS` guard and the delimited reported/normalized values make a truncated capture
obvious, but that is diagnostic hardening, not the repair, and it should not be presented as one.

So this is closed as superseded, not merged. The diagnostic hardening stays available as a small standalone
change if you want a future truncation (or a genuinely wrong path) to be reported as such instead of as an
escaped-deployment-tree failure - say the word and I will open it on its own, otherwise I will leave it.

Kept for the record, no action implied: the assertion still relies on `file(REAL_PATH)` for normalization, which
resolves correctly for an existing anchor path and silently keeps its input when the path does not exist.

The `product_fit.py` half of that diagnosis (`9cc5e20`) is a different story - `#132` is rebased onto the current
head and remains a strict superset of it, because draining the stream continuously also covers the timeout,
banner and exit-code paths and removes the pipe-buffer risk that `process.stdout.read()` leaves in place.

PR #131, opened 2026-09-17
test(product-e2e): capture the QML example's stderr in the E3 product-fit harness

skawu:
## Ruling item 4: the harness observability correction, with an A/B proof on Windows

The Windows GA E3 step kept the QML process alive and captured an empty line list, and you authorised exactly
this change: set `QT_FORCE_STDERR_LOGGING` for the spawned example, keep the offscreen/software rendering
settings, and do not touch product signals or the public QML lifecycle.

**Reproduced locally before the change** (same harness, same example binary, Windows):

```
RuntimeError: qml-basic did not reach initial view-only Running state: []
```

**After the one-line change**, the identical command passes end to end:

```
PASS: qml-basic view-only isolation -> QML stop/configure/start -> control input + client-count reconnect
      -> final stop
```

So the E3 signal was never missing - it was the capture path. That also means the E3 lifecycle itself
(view-only isolation, stop/configure/start, control input, client-count reconnect, final stop) is executed
evidence now, not merely an alive process.

## What changed

```python
env["QT_QPA_PLATFORM"] = "offscreen"      # unchanged
env["QT_QUICK_BACKEND"] = "software"      # unchanged
env["QT_FORCE_STDERR_LOGGING"] = "1"      # added
```

Harness-only: no `console.log` product signal replaced, no second runtime, no QML lifecycle change, no test
weakening or skip.

One file, +5 lines, based on `6169528`. Absorb it as an exact blob if that is easier than merging.


skawu:
Owner disposition after exact-candidate convergence: **do not absorb; close as superseded by the accepted candidate.**

I rechecked current #106 head `4281db23387780d0b246711438d6b8b134a787d3` rather than assuming this one-line harness change had landed. It has **not**: `tests/product-e2e/qml_product_fit.py` on the accepted candidate does not set `QT_FORCE_STDERR_LOGGING`. Nevertheless the exact candidate's integrated V1 GA run `35296682318` is fully green on both reference hosts, including Windows job `105450554009` with `RFB correctness and product E2E — Windows` completed SUCCESS, and #104 is now closed/accepted.

So the historical diagnosis was useful, but this patch is no longer required to make E3 observable or pass. Adding it now would be opportunistic churn on a candidate that has been explicitly frozen for #109 physical/native acceptance.

Disposition:
- no product/QML lifecycle defect remains attached to this PR;
- no further E3 harness change is authorized on the frozen candidate without a new failing execution;
- keep this PR as historical diagnostic evidence only;
- if a future post-V1 or later-candidate Windows logging regression reproduces the empty-capture condition, the same bounded observability change can be reconsidered then.

No effect on #109, compatibility scope, or release authorization.

skawu:
Acknowledged and closing as superseded, exactly as disposed: the accepted candidate passes the relevant executed
surfaces without this one-line change, so there is no defect attached to it and adding it after the freeze would be
opportunistic churn.

One follow-up so the record is accurate: this change had been folded into my structure/docs batch (PR #135) before
your disposition arrived. It is being **removed** from that branch, so #135 no longer carries it, and this PR stays
as historical diagnostic evidence only. If a future candidate reproduces the empty-capture condition, I will reopen
the same bounded observability change against that failing candidate.

Also for the record: my two earlier comments on #109 and the A/B question on #106 were posted through a syntax that
published the file reference instead of the text, so you never saw them. They are being re-posted properly; nothing
was silently pending on your side.

PR #132, opened 2026-09-17
test(qpa): surface the deployed consumer's output while waiting for RFB

skawu:
## Ruling item 3, first bullet: make the clean deployed consumer say what happened

The installed-QPA product-fit could only report a process return code or a socket error when the clean deployed
consumer never became ready, and it did not read the child's pipe at all until after a successful run. That is
exactly why the Linux abort (`consumer exited before RFB listener became ready: -6`, SIGABRT) and the Windows
"alive but never ready" case are opaque, and it also means a chatty child could block once the pipe buffer filled.

## What changed

- the child's output is drained continuously into a buffer by a reader thread while the harness waits for RFB;
- every failure now carries it: early exit, timeout, both banner mismatches, and a non-zero exit code;
- `QT_DEBUG_PLUGINS=1` is set so the captured output shows whether the `hyremote` platform plugin is activated
  at all, and `QT_FORCE_STDERR_LOGGING=1` because a Qt GUI application on Windows logs to the debugger rather
  than the inherited pipe - the same capture-path pitfall already fixed for the E3 harness (#131).

**Boundary respected:** both flags are observability only. The harness still removes every HyRemote/Qt lookup
override (`QT_PLUGIN_PATH`, `QT_QPA_PLATFORM*`, `QML*_IMPORT_PATH`, `LD_LIBRARY_PATH`) before spawning, still
narrows the Windows `PATH`, still runs with the deployment tree as the working directory, the E4 application
source stays Qt-only, and no HyRemote include, call or test hook was added to make anything green.

## Evidence

- local, decisive for this change: pointing `--app` at an executable that cannot accept the harness arguments
  now reports

  ```
  RuntimeError: consumer exited before RFB listener became ready: 2
  unknown option --test-seconds
  usage: ...python.exe [option] ... [arg] ...
  ```

  where the previous harness reported only the return code. The capture and the surfacing both work;
- the hosted Linux SDK job on `6169528` is the case this is for:
  `PASS: 16 HyRemote/Qt dependency resolutions stayed within .../deploy-source-qpa` followed by
  `RuntimeError: consumer exited before RFB listener became ready: -6` with no child output.

## Not claimed

No root cause for the SIGABRT or for the Windows never-ready case. This change exists so the next run names the
missing transition - whether `qhyremote` is activated, whether the native delegate is created, whether a platform
window is observed, whether the composite `RemoteAccess::start()` is reached - instead of hiding it behind a
return code. If you prefer to drop the two debug flags once that transition is proven, say so and I will.

One file, +30/-3 lines, based on the current candidate head; absorb it as an exact blob if that is easier.


skawu:
Rebased onto `eeb3a2e`, so absorbing this is an exact-blob operation; one file, +50/-10.

`9cc5e20` and this PR touch the same file, and this one is a strict superset of it rather than a competing
change - worth reading together:

| | `9cc5e20` | this PR |
| --- | --- | --- |
| early exit | reads `process.stdout` **after** the child is gone | same content, from a buffer drained while waiting |
| timeout | not covered | captured output included |
| banner mismatch | not covered | captured output included |
| non-zero exit code | not covered | captured output included |
| pipe buffer | never drained before success, so a chatty child can block on a full pipe | drained continuously, so that risk is gone |
| plugin/activation evidence | not covered | `QT_DEBUG_PLUGINS=1` shows whether `qhyremote` is activated |
| Windows capture | not covered | `QT_FORCE_STDERR_LOGGING=1`, the same capture-path pitfall as #131 |

The one line that changes behaviour between the two is `process.stdout.read()`: once a reader thread owns the
stream it returns nothing, so this PR removes it rather than leaving it beside the buffer. Everything else is
additive, and the boundary is unchanged - lookup overrides are still removed before spawning, the Windows `PATH`
is still narrowed, the working directory stays inside the deployment tree, the E4 source stays Qt-only, and no
HyRemote hook is added.

Evidence for the mechanism (local): pointing `--app` at an executable that cannot accept the harness arguments
now reports `consumer exited before RFB listener became ready: 2` followed by that child's own stderr lines,
where the previous harness reported only the code.

Still no root cause claimed for the Linux SIGABRT or the Windows never-ready case: the point of this PR is that
the next executed run names the missing transition instead of a return code.

skawu:
Owner disposition after exact-candidate convergence: **do not absorb; close as superseded by the accepted candidate.**

I rechecked the actual file on #106 head `4281db23387780d0b246711438d6b8b134a787d3`. This PR's reader-thread / `QT_DEBUG_PLUGINS` / `QT_FORCE_STDERR_LOGGING` changes are **not present** there. Even so, the exact candidate now passes all relevant executed acceptance surfaces:
- Transparent QPA Proxy: SUCCESS on both reference OSes;
- SDK consumption: SUCCESS on both reference OSes;
- integrated V1 GA run `35296682318`: Linux job `105450553804` SUCCESS and Windows job `105450554009` SUCCESS, including clean installed QPA and combined QML+QPA deployment.

That means the blocker this diagnostic was created to expose has already been resolved on the canonical line, and the current harness is sufficient for the accepted candidate. The proposed continuous drain/debug flags are defensible diagnostic hardening, but they are no longer release-blocking evidence and should not be added after the candidate freeze merely for completeness.

Disposition:
- no merge/absorption into `4281db2`;
- no new QPA diagnostic instrumentation while #109 physical/native acceptance is the critical path;
- preserve this PR as historical diagnostic material;
- if a future execution again fails before the RFB listener or produces an opaque timeout/early-exit, reopen a bounded observability change against that failing candidate rather than modifying this accepted one pre-emptively.

No change to the frozen E4 pure-Qt application rule, deployment contract, or release scope.

skawu:
Acknowledged and closing as superseded, exactly as disposed: the blocker this diagnostic was built to expose is
resolved on the canonical line (Transparent QPA Proxy, SDK consumption and the integrated V1 GA run are all green on
both reference hosts), so the instrumentation is defensible hardening rather than release-blocking evidence, and it
does not belong on a frozen candidate.

Same follow-up as on #131: this change had been folded into my structure/docs batch (PR #135) before your
disposition arrived, and it is being **removed** from that branch, so #135 no longer carries it. If a future
execution again fails before the RFB listener without reporting why, I will reopen a bounded observability change
against that failing candidate.

Issue #134, opened 2026-09-18
test: real-world Qt application compatibility matrix


skawu:
## Goal

Build a maintainable real-world application test suite that validates HyRemote across representative upstream Qt applications, multiple Qt versions, both reference desktop operating systems, and all applicable integration modes **without modifying upstream submodules**.

This work expands test evidence, not product support claims. Passing a third-party application is evidence only unless a compatibility/release authority explicitly promotes a row.

## Final application set

- `qbittorrent/qBittorrent` — L1 complex Widgets / primary pristine-QPA baseline.
- `sqlitebrowser/sqlitebrowser` — L1 lightweight deterministic Widgets / Embedded C++ baseline.
- `musescore/MuseScore` — L2 complex Qt Quick/QML lane.
- `mltframework/shotcut` — L2 mixed Widgets + Qt Quick/QML + multimedia lane.
- `obsproject/obs-studio` — L3 heavy graphics/video/multi-window/native-surface stress lane.

QGIS/Krita remain ad-hoc stress candidates and are intentionally not always-tracked submodules because of disproportionate repository/dependency cost.

## Qt-version policy

Public C++ / QML observation lanes:

- Qt 6.8.0 — minimum public-API floor;
- Qt 6.8.3 exact — V1 reference line;
- latest legally available Qt 6.8 commercial LTS patch (currently recorded as 6.8.9) — licensed observation only;
- Qt 6.9.3 — continuity lane;
- Qt 6.10.3 — forward-compatibility lane;
- Qt 6.11.2 — current stable forward observation;
- Qt 6.12 preview — optional canary only.

QPA uses Qt private ABI:

- V1 qualification remains **Qt 6.8.3 exact only**;
- every other QPA lane must rebuild `qhyremote` against that exact Qt SDK/private headers;
- a successful observation on another Qt patch/minor is not a support claim.

## Upstream cleanliness / reproducibility rules

1. Third-party applications are git submodules under `tests/third_party/apps/`.
2. Tracked submodule worktrees remain pristine; no HyRemote patch/generated/test artifact is committed inside them.
3. Integration lives under `tests/third_party/{specs,overlays,scenarios,harness}` and `tools/third_party/`.
4. QPA uses the pristine application through normal deployment / `-platform hyremote`.
5. Embedded C++ / QML uses a disposable materialized source tree; tracked submodules are never patched.
6. Nested-submodule projects use disposable recursive materialization rather than pretending `git archive` is complete.
[...]

PR #138, opened 2026-09-18
test(realworld): add L2 real builds for Shotcut and MuseScore (#134)

skawu:
Child convergence PR for #134. Base is the real-world test branch, not the V1 product candidate.

This PR upgrades L2 from source-watch/bootstrap to actual upstream build execution:
- Shotcut: pinned pristine tracking commit, upstream Debian unstable dependency recipe, full configure/build smoke.
- MuseScore: pinned pristine recursive source graph, Qt 6.10.3 candidate lane aligned with upstream's current 6.10.x CI family, upstream Linux setup/build entry points.

These jobs first establish that the tracked real-world source can actually build in its applicable Qt/dependency environment. They do not yet claim HyRemote named-app support, and MuseScore/Shotcut QML/QPA behavior remains a subsequent layer after build viability is proven.

No V1 product surface or frozen candidate SHA is modified.


skawu:
### V1 sealing disposition

This qualification work is useful, but it is not part of the V1.0.0.0 release candidate and its parent real-world matrix was already removed from the V1 product path. Keeping an active draft PR during sealing creates noise and invites accidental scope movement.

Close unmerged for V1. Preserve the commits/issue history as post-V1 qualification input under #134; after GA, re-open/re-execute the bounded real-world/Qt-line work against the released/backmerged tree if still valuable.

No support claim or V1 blocker is lost by this closure.

PR #146, opened 2026-09-18
fix(v1): default port 5921 in all three integration modes, user-overridable

skawu:
## Status update 2026-09-19 - rebuilt onto current `develop`

These two commits were replayed onto `develop` @ `64140c8`. The original branch point `eab386e` is 19 commits behind,
and the verification on record belongs to `4112ae6`, so the old evidence no longer describes a head that can be merged:

- `b5b2587` "fix(v1): make 5921 the product default port across all three integration modes"
- `d9d3e57` "feat(v1): make the default listener port configurable at configure time"

The only conflict was **`cmake/HyRemoteProjectOptions.cmake`, and it was purely additive** - develop has since gained the
`HYREMOTE_WITH_TRANSPORT_SECURITY` option and this branch adds `HYREMOTE_DEFAULT_PORT`. **Both are kept**, security
first, with nothing weakened. Diff against `develop`: **24 files, 63 insertions, 37 deletions**.

Verified on the replayed head `d9d3e57` - real numbers, not inherited ones:

- **All ten standalone release-readiness gates PASS** (the parameterized `check_release_profile` runs separately).
- **Configure and build succeed** (Ninja, `C:/Qt/6.8.3/mingw_64`, mingw1310_64, `Release`, `HYREMOTE_BUILD_TESTS=ON`,
  `HYREMOTE_BUILD_EXAMPLES=OFF`); configure reports `Transport security: ON`, i.e. the switch that landed on develop and
  the option this branch adds coexist.
- **`ctest`: 43/43 passed**, in **both directions**: the shipped default, and a build configured with
  `-DHYREMOTE_DEFAULT_PORT=5930` (configure accepted the value, the build succeeded, and the suite stayed green). That is
  what shows the library follows the configuration rather than a literal.
- **Test count differs from the previous record on purpose, and the difference is the configuration**: this run had
  `HYREMOTE_BUILD_QML_API` and `HYREMOTE_WITH_QPA_PROXY` at their defaults (`OFF`), so the QML and QPA test cases were not
  registered. The older 72/72 figure came from a build that enabled those modes. The two numbers are not comparable and
  neither is being presented as the other.
- **Runtime observation taken, and it separates two things the earlier record kept apart.** `hyremote-widgets-basic`
  launched with **no port argument** prints `READY 5921` from the shipped-default build - and **also prints
  `READY 5921` from the `-DHYREMOTE_DEFAULT_PORT=5930` build**. That is not a defect and not news: the shipped examples
  pass their own explicit port default, so an example without `--port` does not inherit the configure-time value; the
  configure-time value governs the integrator case, an application that never calls `setPort()`. Making the examples
  inherit it is a deliberate, separately recorded small follow-up. So:
  - the evidence that the **library default follows the configuration** is the facade-level assertion in the test suite
    (the four default-pinning assertions compare against the macro), which is why the 5930 tree passes its suite - **not**
    an example print;
  - the example print above documents the actual shipped behaviour, and anyone re-running this should expect `READY 5921`
    in both configurations.

Recorded separately, not fixed here: `examples/remote-support-showcase/README.md:35` still illustrates `--port 5901`;
that file is not part of this branch.

## What this changes

Product requirement: **the default port is 5921**, and **all three integration modes let the user change the port**.

### The default moved to 5921 in the two places that own a default

| Where | Before | After |
| --- | --- | --- |
| `src/remoteaccess/src/remote_access.cpp` - Core default listener port (the Embedded C++ API, and the declarative QML API through its editable `port` property) | 5900 | **5921** |
| `integrations/qpa/hyremote_qpa_remote_controller.hpp` - QPA proxy default, used when the platform string carries no `hyremote-port` | 5900 | **5921** |

The QML module needs no separate default: `QmlRemoteAccess.h` declares
`Q_PROPERTY(int port READ port WRITE setPort NOTIFY portChanged)` and `setPort` forwards to `RemoteAccess::setPort()`,
so it inherits the Core default and stays user-editable.

### Changing the port was already possible in every mode, and is now verified by running it

Code reading is not evidence, so each mode was run with an override on this branch:

| Mode | Override mechanism | Observed |
| --- | --- | --- |
| Embedded C++ (`widgets-basic`) | `--port 5930` (`RemoteAccess::setPort()`) | `READY 5930` |
| Declarative QML (`qml-basic`) | `--port 5931` (writable `port` property) | `READY 5931` |
| Transparent QPA proxy | `-platform "hyremote:hyremote-port=5932"` | `HyRemote QPA Proxy active; native delegate: "windows" remote address: "127.0.0.1" port: 5932 remote input: false` |

And the **default** in each mode, with no port argument anywhere and a standard viewer connecting to 5921:

| Mode | Observed |
| --- | --- |
| Embedded C++ (`widgets-basic`) | `READY 5921`, viewer connected, frame captured (7961 B) |
| Declarative QML (`qml-basic`) | `READY 5921`, viewer connected, frame captured (15274 B) |
| Transparent QPA proxy (`-platform hyremote`) | `... port: 5921 ...` and `remote access active on "127.0.0.1" 5921`, viewer frames captured |

### Tests

Four assertions referenced the old default and were updated: `src/remoteaccess/tests/test_remote_access.cpp`,
`integrations/qml/HyRemote/tests/test_qml_module.cpp`, `integrations/qpa/tests/qpa_remote_config_test.cpp`,
`tests/consumer-installed-sdk/main.cpp`. Full suite after a clean rebuild:
**72/72 passed, 0 failed** (50.23 s).

### Documentation and examples follow the product default

`README.md`, `docs/architecture.md`, `docs/guide/install.md`, `docs/en/guide/install.md`,
`docs/getting-started/{cpp,qml,qpa-proxy}.md`, `docs/v1-api-stability.md`, `docs/viewer-connection.md`, the four
example READMEs and the four example programs (23 files, 37 lines). The viewer display notation is corrected with the
port: `5921` is display **`:21`**, where `5900` was `:0`.

### Deliberately not touched

`research/capture/evidence/host-windows-rhi-default-*.json` still contains the digits `5900` - inside recorded latency
measurements (`"guiLatencyP95Ms": 9.905900000000088`, `"p50Ns": 2685900`), not as a port. A blanket replacement would
have silently corrupted recorded data, which is exactly the kind of edit that must never be made to an evidence file.

### Note for whoever lands this

This advances the candidate SHA, so any physical-acceptance evidence that names a tree now needs to name this one.
The Windows E1-E4 sets currently recorded on #109 were taken on `eab386e` and would be re-taken on the new head.


skawu:
## Extended: the default port is now configurable at build time (decision item 10)

The decision round asked for the default port to be definable by an integrator at build time, not only overridable at
run time. Added as a second commit on this branch, because it is the same feature and stacking it separately would
move the candidate twice.

### Mechanism

- `cmake/HyRemoteProjectOptions.cmake` gains `HYREMOTE_DEFAULT_PORT` (cached string, default **5921**) with range
  validation, and publishes it to **every** target through `add_compile_definitions`, so the library, the QPA plugin,
  the examples and the tests cannot disagree about the default.
- The two owners of a default use it now: the Core facade and the QPA proxy configuration. The declarative QML API
  keeps inheriting the Core value through its writable `port` property - no second default to keep in sync.
- The four assertions that pinned `5921` now compare against `HYREMOTE_DEFAULT_PORT`, so a build configured with a
  different default is self-consistent instead of failing its own tests.
- Docs: canonical install guide (Chinese and English) and the C++ getting-started page.

### Verified in both directions

| Configuration | Build | `ctest` | Application with no port argument |
| --- | --- | --- | --- |
| shipped default | ok | **72/72 passed** | `READY 5921` |
| `-DHYREMOTE_DEFAULT_PORT=5930` | ok | **72/72 passed** | example reports `READY 5921` - see below |

The 5930 row is the proof that matters: the suite passes because the assertions and the library agree on the
**configured** value, not on a literal. After the check the tree was restored to the shipped default and the suite was
re-run green.

### What the 5930 row also shows, stated plainly

The shipped examples pass their own explicit port default, so an example launched without `--port` still listens on
5921 even in a 5930 build; `--port` (or the QPA `hyremote-port` parameter) is what moves them. The configure-time
value governs **applications that do not call `setPort()`**, which is the integrator case the option exists for. Making
the examples inherit the configured default rather than pinning 5921 is a small follow-up, deliberately not bundled
here so that this change stays reviewable.

### One known inconsistency, disclosed rather than hidden

`tests/consumer-installed-sdk/main.cpp` still pins the shipped `5921`. It is not part of `ctest` on this host, and
under a custom-port configuration that fixture would need the value published by the installed package (e.g. a
`HyRemote_DEFAULT_PORT` entry in `HyRemoteConfig.cmake`). Recorded as a follow-up rather than silently left for
somebody to trip over.

Issue #143, opened 2026-09-18
[V0.2.0.0][SECURITY] Ship truthful optional access control; keep native encryption non-blocking


skawu:
## Authority

V0.2.0 security/access-control authority under #233.

## Product decision

HyRemote's security/access policy is **optional and default-off**. A user explicitly enables an available policy. Default-off is a product usability decision and is never an Internet-safety claim.

The first user trial must ship a truthful, deterministic security/access state without turning V0.2 into an identity platform or native-TLS infrastructure release.

Architecture may preserve future capability; implementation stays minimal. Sunk implementation does not create product scope.

## User-facing model

The product must describe observable facts rather than claiming to know that a network is "trusted":
- Authentication: on/off/profile actually enabled;
- Transport encryption: on/off/profile actually enabled;
- Remote input: on/off;
- Effective listener endpoint;
- clear warning when an unencrypted/unauthenticated listener is exposed beyond loopback.

Documentation may recommend use only on networks the deployer trusts, but Runtime/diagnostics must not present `Trusted LAN` as something HyRemote can technically prove.

### Security disabled — default

- no authentication/encryption is enabled by default;
- #174 owns the actual listener behavior currently shipped, including the V0.2 wildcard/default and explicit/interface modes already implemented;
- docs/runtime messaging states the actual unauthenticated/unencrypted exposure and **not Internet-safe** boundary;
- no silent address substitution and no hidden security upgrade/downgrade.

### Password authentication — optional baseline capability

Use the simplest maintained-viewer-compatible password authentication available to HyRemote when the user enables it.

Required behavior when shipped/enabled:
- correct credential succeeds;
- wrong/missing credential fails deterministically;
- bounded handshake/timeouts;
- no secrets in logs/events/getters/diagnostics;
- reconnect starts cleanly;
- authentication must not be described as transport confidentiality.

Simple viewer-compatible password authentication is part of the broader GA remote-access baseline under #343, but V0.2 may ship its first trial with the exact mode that is already ready and truthfully qualified.

### Native encrypted transport — optional/differentiating

Existing VeNCrypt/X509Vnc/TLS work may remain implemented/experimental, but:
- it is not a V0.2 customer-trial blocker;
- it is **not an inevitable V1 GA requirement**;
- native encryption competes with VPN/gateway/tunnel/other secure-network solutions and must be admitted as a product capability by evidence/Product Owner decision;
- if exposed, unavailable dependencies/configuration fail closed;
[...]

Issue #33, opened 2026-09-15
[V1.0.0.0][GA] x86 Windows + Linux first-generation product release gate


skawu:
## Product milestone

**V1.0.0.0 — HyRemote first-generation GA on x86_64 Windows + Linux**

This is the final product-level GA authority. Draft PR #106 is the sole V1 integration line; #95 owns release mechanics; #165 owns exact-candidate freeze/change control.

## Frozen product/application architecture

V1 keeps the already accepted three-mode product model:

1. Embedded C++ — `HyRemote::RemoteAccess`;
2. Declarative QML — `import HyRemote`, thin wrapper over the same runtime;
3. Transparent QPA — package-owned `qhyremote`, ordinary applications remain Qt-only.

Architecture invariants:

- one shared `HyRemoteRemoteAccess` normal runtime;
- Core remains STATIC internal/source composition, not an installed application SDK target;
- QML does not create a second Core/Session/transport stack;
- QPA is a native-delegate-preserving proxy and exact-private-ABI payload, not a replacement-only second product;
- one `hyremote_deploy(TARGET ... [QML] [QPA])` deployment entry point;
- backend/platform implementation types stay out of stable application APIs.

A fix that creates multiple product runtimes/SDK personalities merely to satisfy a gate is a V1 design regression.

## Reference release line

Mandatory reference environment:

- Windows x86_64;
- Linux x86_64;
- exact Qt 6.8.3 for the integrated all-three-mode candidate;
- exact Qt 6.8.3 QPA private ABI for the supported reference QPA payload.

Additional Qt lines are governed by #57. Completing qualification in V1 does not automatically promote every line to Supported.

## Required product capabilities

Across the accepted reference combinations:

- remote view;
- remote pointer/keyboard/text path;
- native local display/input coexistence;
- disconnect/reconnect;
- bounded protocol/input/backpressure/client occupancy;
- reproducible build/install/source/deployment;
- consistent lifecycle/diagnostics across the three modes;
- compatibility/security/known-limitations truth backed by exact evidence.

## Security requirement — superseded scope

Older V1 wording that accepted SecurityType None as the final GA security boundary is superseded by the latest x86-first scope ruling (#147/#157).

- #143 authenticated/encrypted transport is now mandatory V1 x86 work.
- Until #143 lands, the **current candidate truth** remains SecurityType None; documentation must not pretend otherwise.
- #153 is an interim non-loopback fail-closed guard and does not close #143.
[...]

PR #153, opened 2026-09-18
security: fail closed on a non-loopback listener until authenticated transport exists

skawu:
Product ruling (2026-09-18): the security posture is **fail-closed** until #143 is fully implemented, and the default loopback listener must be unaffected.

The listener speaks RFB SecurityType None, so a non-loopback bind publishes an unauthenticated remote-control surface. Rather than warn and proceed, RemoteAccess::start() now refuses to start when the listen address is not loopback, sets InvalidConfiguration with a message naming the address and issue #143, and leaves the address unchanged so a later loopback start still works. The default (loopback) path is untouched, which is why the existing suites are expected to stay green.

A unit check was added in tests/test_remote_access.cpp: binding to QHostAddress::Any must fail, the error must be InvalidConfiguration, and the message must say 'without authentication'.

Honest limits of this change:
- #143's two-step implementation still has to deliver the real work; this gate only closes the insecure-by-default risk while it exists.
- No local build or ctest was run on this branch yet - the hosted envelope is the first verification, so I am not claiming local green.
- The test file line it anchors on is also touched by PR #146 (the 5921 default), so whichever lands second resolves one trivial conflict there.

Refs #143, #33


skawu:
This PR is now superseded by the accepted security-profile implementation on current `develop` and should not remain as a competing V1 integration slice.

Current `RemoteAccess::start()` already enforces the stronger product rule: `RemoteSecurityProfile::Insecure` refuses every non-loopback listener, while non-insecure profiles fail closed until their descriptor/capability can actually be honored. That absorbs #153's interim guard into the final #143/#170 profile model rather than preserving a parallel boolean-era patch.

Do not merge this stale branch onto current develop. The active security convergence is #143/#170, with #192/#193 accepted and #196 handling the build dependency contract.

Refs #143 #170 #192 #193 #196

Issue #148, opened 2026-09-18
[POST-V1][EMBEDDED] Transparent QPA Proxy has no EGLFS delegate: scope statement, guide correction and roadmap decision


skawu:
## Goal

Make the embedded applicability of **Transparent QPA Proxy** explicit and decided, so a downstream integrator does not have to read `integrations/qpa` to discover that the mode cannot be used on an EGLFS-class target:

1. document the qualified native delegate set and state that an EGLFS/Wayland-class Linux target has **no delegate** for the QPA payload (i.e. `-platform hyremote` cannot start the application there, not merely "unverified");
2. stop presenting QPA as the embedded default in the cross-compilation guide for targets that have no qualified delegate;
3. record a scope/roadmap decision on a QEGLFS-style delegate (planned, or explicitly out of scope);
4. answer the embedded deployment question (application-local plugin directory; cross-compile-friendly payload list).

Reported from a downstream product integration (Huayan HMI, Qt 6.8.3 / RK3588 / EGLFS), verified against commit **4112ae6**. Not a V1.0.0.0 GA ask; we are **not** blocked — we ship Embedded C++ mode (see below).

## Product context

| Item | Value |
|---|---|
| Milestone | POST-V1 (found while cross-compiling a downstream product against the V1 candidate line) |
| Integration mode(s) | Transparent QPA (subject of this request); Embedded C++ (the mode we actually adopted) |
| Layer(s) | `integrations` (qpa), packaging/deployment, docs |
| Application types | `QQuickWindow` — one fullscreen 1920x1080 window, frameless; Widgets/Quick3D composition not used by the consumer |
| Public API impact | none |
| Qt/private API impact | no new private API requested; the finding is about the existing private-ABI QPA payload |
| Platform-specific impact | Embedded Linux / RK3588 (aarch64) / **EGLFS + OpenGL ES 3.0**; Qt supplied by the BSP/rootfs; application deployed as one directory, runs non-root |
| Affected product milestone/version | post-V1 (`V0.0.3.0` QPA scope is x86 by decision — see #32) |

### 1. On Linux the proxy delegates to `xcb` only

- `integrations/qpa/hyremote_platform_plugin.cpp:31-40` — `referenceNativeDelegate()` returns `"windows"` on Windows and `"xcb"` on Linux; there is no other candidate.
[...]

Issue #141, opened 2026-09-18
[POST-V1] Deferred backlog: audit items, review findings, docs gaps, branch governance


skawu:
## Purpose

This issue is now restricted to **genuinely deferred / post-V1 work**. During the 2026-09-18 systematic review, several entries previously parked here were reclassified as V1 because they are x86-testable under the latest scope ruling (#147/#157). Those items now have dedicated owners and must not be treated as deferred merely because they were once listed here.

Project retrospective: #161. Current handoff: #156.

## Migrated out of this backlog — active V1 work

The following findings are no longer owned by this post-V1 backlog:

- Core unlocked component lifetime race + post-stop statistics -> **#159**.
- RFB 3.7/3.8 protocol truth + bounded Normal-phase client occupancy -> **#158**.
- Widgets/Quick routing, non-1 DPR and IME coverage -> **#162**.
- QPA exact runtime Qt identity, native-delegate controlled failure, HyRemote parameter validation -> **#163**.
- callback exception containment, `noexcept` audit, CompositeInputSink semantics, bounded Quick empty-grab retry, CPU-frame stride validation and private export hardening -> **#164**.
- release-candidate churn / exact-SHA evidence control -> **#165**.
- global V1/post-V1 scope authority mismatch -> **#157**.

Do not duplicate fixes here. Close evidence on the dedicated issues.

## Already fixed / superseded facts

- The high-risk findings fixed by #139 remain fixed: RFB client UAF, composite recoverability downgrade, silent “supported” claim and missing native-delegate assertion.
- GitHub `delete_branch_on_merge` is now enabled; the older statement here that it remained disabled is stale and is withdrawn.
- Historical branch accumulation was reduced substantially. Current residual branch cleanup is tracked operationally in #156 and must use SHA/content checks before deletion.
- Documentation-zone reorganization is no longer merely a deferred idea: PR #155 actively implements physical `docs/reference` / `docs/acceptance` zones and associated gates. Do not create a parallel docs-migration task here.
- `.gitignore` Python byte-code hygiene is being handled by PR #150; do not retain it here as post-V1 work if that PR lands.

### Naming / source-tree conventions

- Codify intentional naming/header conventions in a dedicated naming-conventions document if still useful after V1:
  - Core public headers under `hyremote/core/*.hpp`;
  - product facade under `HyRemote/RemoteAccess.h`;
  - QML URI directory `integrations/qml/HyRemote`;
  - QPA integration naming.
[...]

PR #135, opened 2026-09-18
chore(v1): structure, naming and layout batch from the audit follow-up

skawu:
Draft: the batch agreed with the product side after the #133 audit. Based on `4281db2`.

## Landed so far

| Commit | Content |
| --- | --- |
| `6e9a4c5` | `chore: ignore generic Python and cache artifacts` - `__pycache__/`, `*.py[cod]`, `.cache/`. Tool-specific working directories deliberately stay local/global excludes rather than repository rules. |
| `2d22d18` | `docs(structure): codify naming conventions and forward-compatible layout rules` - new `docs/naming-conventions.md` (the four intentional spellings and where each applies, C++ casing, the two intentional public-header forms, the QML-URI directory rule, the prefix rule at global boundaries, and an explicit list of what V1 does not rename); `docs/repository-layout.md` gains the extension rules that keep later work additive - `src/` grows by product module, `integrations/` grows one directory per payload, a future Qt LTS line gets `integrations/qpa-<qt-line>/` instead of widening the qualified 6.8.3 payload, `tests/` grows by test scope, recorded evidence grows under `docs/acceptance/`, `examples/` grows by integration mode and usage scenario, and a new top-level directory is a structural decision with migration-level authority. |
| `b84fac6` | `test(release-readiness): enforce the payload contract for every integration payload` - the three payload rules now apply to every payload under `integrations/`, not only QML and QPA, so a future payload cannot escape the shared-runtime contract; `docs/naming-conventions.md` became a required document. |

Two conventions from the product side are now recorded in the layout document: **`tests/` holds unit tests and
integration test cases**, and **user-facing usage examples - including examples that combine HyRemote with
third-party open-source applications - live under `examples/`**. Recorded review/acceptance evidence is therefore
documentation and belongs under `docs/acceptance/`, not in `tests/`.

Also corrected: the layout tree said "product examples E1-E5" while the accepted suite is E1-E6 (the
installed/source SDK consumers are E6).

## Still coming in this batch

- the five example README gaps (troubleshooting everywhere, deployment for the two Embedded C++ basics,
  Windows/Linux and build for the showcase);
- an opt-in example that combines HyRemote with a third-party open-source application - fetched at a recorded
  commit into an ignored build directory, OFF by default, outside the default build and the acceptance graph, with
  no third-party sources vendored into this repository;
- folding #131 and #132 (test-surface only) so they do not need candidate/envelope slots of their own;
- SPDX identifiers on the C++/QML sources as one isolated mechanical commit.

[...]

PR #155, opened 2026-09-18
docs: physical documentation zones, machine-checked guide mirrors and the QPA delegate scope statement

skawu:
## What this is

Two documentation-correctness changes that share the same file set, so they land together.

### 1. The declared documentation zones are now physical (ADR-0005)

`CONTRIBUTING.md` and `docs/README.md` already declared three documentation zones and a bilingual-mirror duty, and `docs/repository-layout.md` already declared `docs/acceptance/` as the home of recorded evidence - but `docs/reference/` and `docs/acceptance/` did not exist, 50 documents sat flat in `docs/`, and the only documentation gate checked tokens *inside* documents, never the zones or the mirrors.

- `docs/reference/` created and the 16 product final-state contracts moved into it (`docs/` flat count 50 -> 19); `docs/acceptance/` created with a record contract;
- **language rule amended** (product ruling): the reference zone is **English-canonical** with an optional Chinese mirror, and the mandatory bilingual-pair duty is scoped to `guide/`;
- zone indexes written (`docs/reference/README.md`, `docs/acceptance/README.md`);
- references updated in the same change as required: two workflow path filters, toolchain CMake comments, `src/core` header comments, and the release-readiness checks that open documents by path;
- **the gate is extended and verified**: `check_release_documentation_layout.cmake` now fails when the zones are missing, when a contract sits flat or is absent from the reference zone, when a `docs/guide/*.md` loses its English mirror, or when the language rule, evidence boundary or decision record is edited away. Run standalone with `cmake -P`; it failed first on a real inconsistency this migration introduced and passes now.

### 2. The Transparent QPA scope statement (#148, V1 half)

An EGLFS/Wayland-class Linux target has **no QPA delegate**, so `-platform hyremote` cannot start an application there - that is *not available*, not *unverified*. Both cross-compilation guides now name the qualified delegate set (`qwindows`, `qxcb`), stop presenting QPA as the embedded default, document the deployment contract (application-local `plugins/platforms/`, the helper's payload list, fail-closed exact-Qt check), and state the startup-only boundary with the Embedded C++/QML route. `docs/compatibility.md` carries a separate QPA row, ADR-0004 records the decision (a QEGLFS/Wayland-style delegate is planned with the `V1.x` embedded expansion), and the roadmap says so. The mode-switch recovery command both guides printed - which the script rejects because the mode guard runs before the clean - is corrected, and the residual is tracked on #141.

## Scope and honesty

- Documentation, CI path filters, comments and the gate only; **no product code, no build default and no artifact path changes** - the V1 layout freeze is respected.
- Part 2 is **not a V1 gate** and is not progress on any `V1.x` track; it corrects a present-tense false claim.
- Neither part authorizes anything: acceptance remains with the tracking issues.

Refs #148 #135 #141


skawu:
### V1 sealing disposition: defer repository-wide documentation migration

This PR mixes a broad documentation-zone migration (moving many contracts, changing path filters/gates and repository structure) with a truthful QPA scope correction whose core product statement is already governed elsewhere: V1 QPA is Windows `qwindows` / Linux `qxcb`, exact Qt 6.8.3, while EGLFS/Wayland-class embedded delegates are post-V1.

The structural migration is not required to close a current V1 blocker and creates high path/gate churn during sealing. We should not move dozens of release-authority documents merely to improve repository taxonomy immediately before RC freeze.

Disposition:
- do not merge the repository-wide zone migration into V1;
- preserve ADR/documentation-zone work for post-GA cleanup;
- any present-tense false compatibility statement that still exists on `develop` must be corrected as a **small targeted documentation patch** under #41/#57/#163, without moving the documentation tree;
- no QPA support expansion is authorized.

Closing this combined PR as deferred scope.

Issue #170, opened 2026-09-19
[V1.0.0.0][PRODUCT-SECURITY] Freeze secure configuration and authenticated-session operations across C++/QML/QPA


skawu:
## Product finding

V1 scope changed after the original public API freeze: #143 authenticated + encrypted transport is now mandatory for V1.0.0.0, while the already-closed #101 API freeze only froze listener address/port, remote-input policy, lifecycle/state/error and aggregate `connectedClientCount()`.

That surface is insufficient to productize #143 safely. In particular, the current frozen contract has no complete application-facing answer for:

- authentication/encryption profile selection;
- credential / certificate / key provisioning without secrets in command lines, logs or diagnostics;
- authenticated-session identity distinct from a raw socket/backend client;
- session connect/authenticate/reject/disconnect audit events;
- operator-initiated termination of a connected authenticated session;
- bounded admission/concurrency policy once authenticated sessions exist;
- equivalent secure configuration for Embedded C++, QML and zero-code QPA without creating a second runtime or leaking RFB/OpenSSL/backend types.

#91 intentionally excluded client identity, per-client permissions and session lists because the V1 transport was unauthenticated at that time. #143 now changes that premise. This issue closes the resulting **product-surface gap**; it is not a request to build an identity/account platform.

## Relationship to authorities

- **Child / required product-surface closure of #143.** #143 must not close while this issue is unresolved.
- #101 must be re-audited/re-accepted after the final additive security/session surface is known; its previous closure predates the V1 security promotion.
- #107 owns final package/notices/release metadata, including OpenSSL or any other shipped security dependency.
- #165 still owns RC-FROZEN; none of this may be invented after candidate freeze.

### 1. One transport-neutral secure profile

Define the smallest stable application-facing configuration that can express the accepted #143 secure deployment profile without exposing `RFB SecurityType`, OpenSSL handles, sockets or backend objects.

The contract must distinguish at least:

- insecure/legacy mode where explicitly retained for loopback/trusted test use;
- authenticated mode;
- authenticated + encrypted mode required by the final V1 secure profile.

Do not make transport/backend selection a normal application responsibility.

### 2. Secret provisioning

For all supported modes define a secure, deployable way to supply password/credential/certificate/private-key material.

Requirements:

- no password/private key in normal command-line arguments;
- no secret in logs, `lastError()`, audit messages or crash diagnostics;
- no secret copied into QML-readable diagnostics;
- lifetime/ownership and clearing behavior documented;
[...]

Issue #165, opened 2026-09-18
[V1.0.0.0][RELEASE] Establish exact-candidate freeze and post-freeze change-control before physical acceptance


skawu:
## Problem

The current V1 process correctly requires #109 physical evidence to name an exact candidate SHA, but there is no equally explicit **candidate-freeze transition** that prevents routine V1 changes from continuing after physical evidence starts.

This has already caused rework: physical evidence was recorded on one #106 SHA while later V1 PRs changed product defaults, security behavior, documentation contracts and example binaries (#146/#153/#155/#160). Even when those changes are individually correct, evidence against the earlier tree is no longer evidence for the later candidate.

The defect is release-process structure, not any individual contributor or PR.

## Required lifecycle

Introduce an explicit state between repository convergence and physical acceptance:

1. **CONVERGING** — V1-targeting product/docs/test PRs may still land.
2. **RC-FROZEN** — one exact #106/release-candidate SHA is declared. All mandatory hosted/reference gates pass on that SHA; final #109 evidence may begin.
3. **PHYSICAL-ACCEPTANCE** — only evidence collection is expected. Any code/product/package/example change must go through post-freeze change control.
4. **ACCEPTED** — #33 authorizes release flow on the same accepted tree.

## Post-freeze change control

For every proposed change after RC-FROZEN, require an explicit disposition:

- **A — evidence-neutral:** prose/comment-only and machine-proven not to alter built/package/example payloads used by acceptance. May land without invalidating unrelated physical rows, but must still record why.
- **B — partial re-test:** changes one acceptance payload/behavior; list exactly which hosted/physical cells are invalidated and must be re-run.
- **C — full candidate reset:** product/runtime/package/default/security/QPA ABI or shared example change; freeze is revoked, hosted GA re-runs, then #109 restarts against the new SHA.
- **D — defer:** valuable but not required for the current release; hold for post-release/backmerge rather than churn the candidate.

No PR may self-declare itself evidence-neutral merely because it is “small.” The impact is determined by built artifacts and acceptance cells.

## Machine/repository enforcement

Where practical:

- maintain one canonical file or generated record containing current candidate state + exact SHA + freeze timestamp;
- release-readiness checks must refuse “physical acceptance complete” metadata when the evidence SHA differs from the frozen candidate;
- #109 template/results must reference the frozen SHA;
- any merge to the frozen candidate branch after freeze must require an acceptance-impact field (`neutral`, `partial`, `reset`, `defer`) and the affected cells;
[...]

Issue #175, opened 2026-09-19
[V1.0.0.0][INTEROP] Qualify maintained VNC viewers against the final secure/compressed profile


skawu:
## Product gap

V1 currently requires “a maintained standard VNC/RFB viewer” in multiple acceptance issues, but does not own one final viewer-compatibility authority. That was tolerable for the Raw/SecurityType-None correctness baseline; it is no longer sufficient once #143 introduces authentication/encryption and #144 introduces negotiated compression.

A server can pass protocol/unit tests and still be unusable with the viewer an external developer actually installs. V1 therefore needs a small, named, evidence-backed viewer matrix. This is **qualification**, not a promise that every VNC viewer is supported.

## V1 goal

Select and qualify at least one maintained GUI viewer usable for each reference OS against the final V1 profile. The same cross-platform viewer may satisfy both rows if it is actively maintained and the exact version is recorded.

The qualified matrix must cover the final combinations actually shipped:

- RFB 3.8 handshake;
- Raw fallback;
- the selected #144 compressed encoding;
- the selected #143 authentication/encryption mechanism;
- view-only session;
- pointer/keyboard/text control when enabled;
- framebuffer resize/desktop-size behavior used by HyRemote;
- disconnect/reconnect;
- multi-client behavior if V1 documents it;
- failure behavior for unsupported/insecure profile combinations.

## Selection criteria

A V1-qualified viewer must be:

- maintained and obtainable through a normal user path;
- usable without HyRemote-specific patches;
- license/distribution neutral from HyRemote's perspective (the viewer need not ship inside HyRemote);
- capable of the chosen #143 secure profile and #144 encoding, or the server profile must be adjusted before freeze;
- version-pinned in evidence because viewer behavior changes across releases.

`vncdotool`/test clients remain useful automation but do not by themselves satisfy the GUI-viewer product qualification.

## Required matrix

For Windows x86_64 and Linux x86_64 record:

- viewer product + exact version;
- download/source identity;
- HyRemote RC-FROZEN SHA;
- Qt/OS reference row;
- secure profile negotiation result;
- Raw fallback result;
- compressed encoding result;
- framebuffer update/resize result;
- pointer/buttons/wheel/keyboard/basic text result;
- reconnect result;
- view-only rejection of remote input;
- known viewer-specific limitations.

[...]

Issue #107, opened 2026-09-16
[V1.0.0.0][RELEASE-READINESS] Complete release notes, notices and package manifest


skawu:
## Goal

Repository-owned release-readiness authority under #33/#95. It freezes truthful release notes, notices, package manifest, compatibility/security state and machine checks for the exact V1 candidate. It does not create a second product architecture.

Draft PR #106 is the single V1 integration line; #157 owns scope convergence; #165 owns exact-candidate freeze.

## Current state

The repository already contains candidate release notes, NOTICE, package manifest, installed license/notice rules and deterministic release-readiness checks. They remain **candidate** evidence and must now be updated to the latest mandatory V1 set before they can authorize anything.

## Release notes requirements

Canonical `docs/releases/v1.0.0.0.md` must describe only capabilities actually present in the accepted candidate and must remain `candidate / acceptance pending` until #33.

It must include:

- all three integration modes and the one-shared-runtime artifact model;
- Windows/Linux x86 reference environment and exact Qt 6.8.3 QPA line;
- exact compatibility dispositions from #57;
- final security behavior from #143 (not the historical SecurityType None limitation once #143 lands);
- final encoding/bandwidth behavior from #144/#9;
- known limitations and genuinely post-V1 embedded/hardware work;
- exact candidate/evidence identity from #165/#109.

Do not pre-document future capabilities. Until a required feature lands, current documentation must remain truthful about the current candidate.

## Package / artifact manifest

Preserve the frozen product shape:

- one normal application C++ target `HyRemote::RemoteAccess` backed by shared `HyRemoteRemoteAccess`;
- Core STATIC/internal/source only, not installed/exported as a normal app target;
- optional QML payload over the same runtime;
- exact-Qt `qhyremote` platform MODULE payload, no application `HyRemote::QpaPlatform` target;
- one `hyremote_deploy(TARGET ... [QML] [QPA])` deployment entry point;
- license/notices/docs metadata;
- transport/security/encoding implementation names remain private unless their user-visible configuration is explicitly part of the accepted public contract.

## Dependency / notice policy

Distinguish:

- HyRemote Apache-2.0 source/package;
- Qt as external build/runtime/deployment dependency with adopter-specific licensing obligations;
- test/CI tooling that is not automatically product payload;
- any security/compression library actually bundled/copied by #143/#144 must be deliberately added to package/license/notice checks before GA.

## Authoritative V1 mandatory set

Release-readiness must machine-check the same set as #33/#95/#157:

[...]

PR #185, opened 2026-09-19
feat(security): #143 authentication configuration surface with a fail-closed start

skawu:
## What this is

The first implementation increment of **#143 S2** (authentication). It is **configuration only**, and deliberately cannot
pretend otherwise.

The design (`docs/security-model.md` section 10.1, merged via PR #169) freezes RFB VNC authentication (security type 2,
including the 3.8 challenge/response and its blinding) and states that a configured authentication mode must **never be
downgraded** to an unauthenticated `SecurityType None` stream. Since the authenticated transport step is not implemented
yet, the honest configuration surface is one that **refuses to start** rather than a switch that silently does nothing.

## Added

- `RemoteAccess::authenticationEnabled()` and `setAuthenticationEnabled(bool)`.
- `RemoteAccess::setPassword(const QString &)` - an empty string clears it.
- `RemoteAccessErrorCode::AuthenticationUnavailable` (name chosen by the implementer; renamable).
- `start()` checks the authentication state **before composing any component**, so a refused start opens no listener and
  no transport is created. The error distinguishes "authentication was enabled without a password" from "authentication
  was enabled but the authenticated transport is not available yet".
- Setters follow the existing `isConfigurable()` pattern: configuration is mutable only while `Stopped`.
- The password never appears in an error, diagnostic or log line, and is never read back through any property.

## Deliberately not in this change

- **The OpenSSL dependency.** The design requires it to land **with** the capability, never in advance, together with the
  `package`/`LICENSE`/`NOTICE`/deployed-payload updates.
- **Any gate or document change.** `SecurityType None` is pinned by `check_release_metadata.cmake` and the release-policy
  workflow and stated in roughly twenty user documents; those move in the change that lands the capability. Nothing here
  claims authentication exists.
- **Lifting #153's non-loopback fail-closed guard.** The design allows a non-loopback bind only once an authentication mode
  is actually enabled, which this change does not provide.
- **QML/QPA parity** - its own increment. QML will expose an **invokable** `setPassword` plus a read-only
  `authenticationEnabled` (a readable property would let QML bindings read the password back), and the QPA plugin will at
  first support only enabling/disabling, because a password on the process command line is visible to other processes.

## Verification

- `hyremote-remoteaccess-test` passes, including the new assertions: authentication defaults to disabled; enabling it
  makes `start()` fail with `AuthenticationUnavailable` while the state stays `Stopped` and **`transportFactoryCalls` is
  still 0** (proving no listener was opened); the error for a missing password differs from the error for an unavailable
  transport; the error message does **not** contain the configured password; `setAuthenticationEnabled`/`setPassword` are
  rejected while running; disabling authentication restores a normal start; and configuration is accepted again once
  stopped.
- Full suite **47/47**; **all ten** standalone release-readiness gates pass on this tree (the public header changed, so
  they were re-run rather than assumed).
- The tests use the existing fake runtime, so they need no port and open no listener.

## Release process

Public API is added and no existing behaviour changes when authentication stays at its default (off). Requested
classification under #165: it is more than "A - evidence-neutral" because the built payload gains a public symbol, so
**B - partial re-test** is the likely class - but that is the owner's call. No physical cell is invalidated today because
none has been taken on any candidate.

Refs #143 #170 #153 #165 #109

## Follow-up commit: QML and QPA parity

- **QML**: read-only `authenticationEnabled`, plus `Q_INVOKABLE setAuthenticationEnabled(bool)` and
  `setPassword(QString)`. The password is deliberately **not** a property - a QML binding must never be able to read it
  back - and the test asserts that no such property exists. Both invokables return `false` with a local
  `InvalidConfiguration` error while the runtime is not `Stopped`.
- **QPA**: `RemoteConfig::authenticationEnabled`, parsed from `hyremote-auth` with the same boolean vocabulary as
  `hyremote-input` (`0/1`, `false/true`, `off/on`, `no/yes`); an unparsable or value-less parameter fails closed. A
  password is deliberately not accepted here, because `-platform` values are visible in the process command line to other
  processes on the same machine - the host application supplies it through the API instead.

**Verified in a full-mode configuration** (`HYREMOTE_BUILD_QML_API=ON`, `HYREMOTE_WITH_QPA_PROXY=ON`, tests and examples
ON): **75/75** tests pass, including `hyremote-qml-module-test` and `hyremote-qpa-remote-config-test`, and all ten
standalone release-readiness gates pass on this tree. (Test counts are configuration-dependent: 47 in the minimal
configuration, 75 here.)


skawu:
## Architecture review — do not merge this exact API shape

The fail-closed behavior and secret non-echo tests are good, and this HEAD is hosted-green. However #170 predates this PR and requires the final V1 surface to distinguish `Insecure / Authenticated / AuthenticatedEncrypted` and to provide zero-code QPA secret provisioning plus authenticated-session operations.

Two corrections are required before these public symbols enter `develop`:

1. replace the boolean-only `authenticationEnabled` product model with the frozen transport-neutral security-profile model recorded in #170 comment 5740108613; otherwise we would immediately need a second encryption boolean / duplicate state after merge;
2. do not leave QPA in the state `hyremote-auth=on` + no zero-code secret source. QPA may pass a non-secret descriptor path at process start, but raw password/private-key values are never platform parameters.

The useful part of #153 (non-loopback fail-closed until a configured secure profile is real) will be absorbed here rather than merging #153's currently failing old-head branch.

I am treating #185 as the active #143/#170 implementation branch and will re-run the full hosted matrix after the reshape. This is an API/product correction, not a request to broaden scope or create a second runtime.

skawu:
V1 security/product review: **do not merge this increment yet**.

The current QPA answer is not compatible with #170's zero-code secure provisioning contract: `hyremote-auth` can enable auth, but the PR explicitly leaves password provisioning to application API. A transparent `-platform hyremote` adopter has no HyRemote application call site by definition, so secure QPA startup would be enabled-but-unprovisionable rather than a complete V1 mode.

Required before this public surface lands:

1. Freeze one transport-neutral secure profile model covering insecure(loopback/test), authenticated, and authenticated+encrypted modes without backend/OpenSSL types.
2. Define a non-command-line secret provisioning mechanism usable by C++/QML and **zero-code QPA** (for example an explicit secure-profile/config reference whose file/permission/failure semantics are machine-tested). Password/private key material must not appear in `-platform` parameters.
3. Add the authenticated-session identity/lifecycle/termination/admission semantics required by #170, keeping `connectedClientCount()` diagnostic-only.
4. Keep QML as a thin view over the same runtime/session truth; for QPA, define the supported operational mechanism where application callbacks are unavailable rather than inventing a second runtime/API.
5. Only after the complete additive surface is known should #101 be re-audited/re-accepted.

The fail-closed work already present is directionally correct and should be retained. The reason to hold the merge is product/API sequencing: landing an incomplete public security API now would force another redesign before RC freeze.

Refs #143 #170 #101 #107 #165.

skawu:
Follow-up product/architecture review after the branch reshaping: the code now correctly uses `RemoteSecurityProfile { Insecure, Authenticated, AuthenticatedEncrypted }` plus a non-secret `securityConfigFile` path across C++/QML/QPA, and secure profiles fail closed rather than downgrading. Keep those changes.

However, **the PR body is stale** (it still documents `authenticationEnabled + setPassword` and QPA without zero-code secret provisioning), and this branch is still not merge-ready as the V1 security surface because #170 remains only half implemented.

Before this PR (or a successor security PR) can enter `develop`, add the remaining frozen surface and implementation in one coherent security line:
- runtime-owned authenticated session id/identity/lifecycle records; failed auth must not obtain authenticated identity;
- backend-neutral selected-session termination with held-input cleanup exactly once;
- documented bounded default + application-lowerable authenticated-session admission limit;
- QML mirror over the same runtime session truth, no second model;
- explicit QPA operational/audit mechanism while retaining a Qt-only host app;
- descriptor parsing/validation and actual #143 authentication + encrypted transport; secure profiles must stop failing merely as “not available yet” and become real before the public API is frozen;
- Windows/Linux deterministic security evidence and public header/QML/package freeze tests;
- refresh this PR body to describe the actual branch before any merge decision.

Until those are present, keep the branch open and **do not merge the additive public security API piecemeal**. Refs #143 #170 #101 #175 #165.

Issue #164, opened 2026-09-18
[V1.0.0.0][RUNTIME] Contain callback exceptions and close remaining low-level lifecycle hazards


skawu:
## Review finding

#141 still contains several runtime correctness/hardening items that are too concrete to remain buried in a mixed post-V1 backlog. They are x86-testable and should be resolved or explicitly dispositioned before V1 GA under the latest scope ruling.

The remaining items are:

- Core callback entry points (`onFrameReady`, `onCaptureEvent`, `onTransportEvent`) do not contain exceptions; allocation or adapter exceptions may escape into producer callback stacks.
- `RemoteAccess::stop()` is `noexcept` while parts of its cleanup/diagnostic path may allocate.
- `CompositeInputSink::post` may drop an event and then throw when a child reported a deferred error, creating ambiguous exactly-once semantics.
- Quick empty-grab retry can continue without an explicit bounded retry budget.
- `writableCpuFrameOf` derives stride from geometry rather than validating/storing the authoritative buffer stride contract.
- private seam/factory symbols are exported more broadly than needed.

### Exception boundary
- Define which internal callbacks are exception boundaries and guarantee that no exception escapes into Qt/adapter/transport callback stacks.
- Convert failures to the existing HyRemote error/event model without double-reporting or resurrecting an acknowledged recoverable occurrence.
- Add failure-injection tests for allocation/adapter exceptions where practical.

### `noexcept` correctness
- Audit every `noexcept` public/private function that performs allocation, formatting, container growth or callback dispatch.
- Either make the implementation genuinely non-throwing or remove the incorrect `noexcept` where the public ABI contract permits it.
- Never allow `std::terminate` as an undocumented normal failure mode.

### Input exactly-once semantics
- Specify whether `CompositeInputSink::post` accepts or rejects an event before child delivery.
- If a deferred child error exists, do not both partially deliver and then report rejection ambiguously.
- Add deterministic multi-child failure tests.

### Capture retry/resource bounds
- Bound empty Quick-grab retries and expose the terminal/recoverable outcome through existing diagnostics.
- Validate CPU frame stride/buffer-size invariants explicitly before writable access.

### Symbol surface
- Audit exported symbols and hide private factories/seams that are not part of the frozen 1.x application ABI.
- Add an export-surface check if feasible rather than relying only on source convention.

## Acceptance criteria

1. No internal exception escapes through Core/Qt/transport callback boundaries in deterministic failure-injection tests.
[...]

Issue #159, opened 2026-09-18
[V1.0.0.0][CORE] Close concurrent start/stop component lifetime race and preserve run statistics


skawu:
## Review finding

Two Core correctness defects from #141 still reproduce in current #106 HEAD `eab386ef8d41e618c11b94166b7847fb9d2bdb0f`.

### 1. Component lifetime across concurrent start/stop/setter

`Session::start()` captures raw `CaptureSource*` / `Transport*` from the owning `unique_ptr`s and then performs component calls outside `Impl::mutex`. Concurrent `stop()` may transition the Session back to Stopped, after which a public setter is allowed to replace/destroy that owned component while the original `start()` call still holds the raw pointer.

The existing run-generation / teardown-claim logic prevents double teardown, but it does not itself pin the object lifetime across every unlocked component call.

### 2. Cumulative statistics lost at teardown

`SessionStats` is documented as per-run cumulative counters reset by `start()`. Several mailbox-owned values (`framesDroppedByPolicy`, `framesRejectedOverflow`, `lastFrameId`, `maxMailboxWaitingObserved`, `maxMailboxOwnedObserved`) are only overlaid in `stats()` while `impl.mailbox` exists. Teardown resets `impl.mailbox`, so querying stats after stop silently returns zero/default values for events that occurred during that run.

That contradicts the public Core header contract and weakens post-stop diagnostics/acceptance evidence.

### Lifetime correctness

- Define one ownership strategy that guarantees `CaptureSource`, `Transport` and any Core-owned runtime object remains alive for every unlocked call made by `start()`, scheduler/dispatcher and teardown.
- Preserve the existing API rule that component setters are accepted only in Stopped state.
- Preserve concurrent `start()` / `stop()` cancellation semantics; do not solve this by forbidding the already-documented concurrency.
- Do not hold the global Core mutex across arbitrary adapter calls merely to hide the race if that introduces deadlock/reentrancy risk.
- Add deterministic concurrency tests that force the destructive setter interleaving rather than relying on stress probability.

### Statistics correctness

- Snapshot/fold mailbox counters into `impl.stats` before the mailbox is destroyed, or otherwise keep their per-run cumulative values observable after stop.
- Gauges such as current waiting/in-flight may legitimately become zero after stop; historical maxima/counters/last assigned frame ID must retain their run value until the next `start()` reset.
- Add stop/fault/cancel test coverage and document exactly which fields are gauges vs cumulative-per-run values.

## Acceptance criteria

[...]

Issue #157, opened 2026-09-18
[V1.0.0.0][GOVERNANCE] Converge latest x86-first scope ruling across release authorities and gates


skawu:
## Problem

The repository currently contains two incompatible scope narratives:

1. the older integrated V1 candidate and several issue bodies say security/authentication, bandwidth/acceleration, additional Qt qualification and related work are post-V1; and
2. PR #147 records a newer product ruling: **anything that can be completed and validated on x86 must be completed in V1.0.0.0; only work that inherently requires an embedded target/platform may remain after V1**.

This is now a release-governance defect, not merely stale wording. A release owner cannot tell from #33/#106/#107/#141/#143/#144/#57/#9 alone which items are mandatory before GA.

## Latest ruling to apply

Use #147 as the latest scope decision unless a newer explicit owner ruling supersedes it:

- #143 authenticated/encrypted transport: x86-doable -> V1.0.0.0;
- #144 RFB encoding/bandwidth strategy: x86-doable portion -> V1.0.0.0;
- #9 damage-aware / asynchronous acceleration work that is x86-doable -> V1.0.0.0;
- #57 Qt 5.15 / additional desktop LTS qualification work that is x86-doable -> V1.0.0.0;
- #17 DMA-BUF/GBM/external-buffer work requiring embedded graphics stack -> post-V1;
- #10 embedded hardware encode validation -> post-V1;
- additional embedded platform delegates/targets -> post-V1.

This issue does not choose a new scope. It propagates an already-recorded ruling into all authorities and gates.

## Required work

1. Audit and update #33, #106, #107, #95 and the V1 release notes/roadmap so their mandatory dependency set matches the latest ruling.
2. Update titles/bodies/labels of #143/#144/#57/#9 where they still say `POST-V1` or otherwise contradict the ruling.
3. Reclassify #141: split or move every x86-doable correctness/security/performance item that is now V1 work; leave only genuinely deferred/embedded/non-blocking work under a post-V1 heading.
4. Ensure `.github`/release-readiness gates encode the same scope where machine enforcement is possible. A document saying 'V1' while the release gate ignores it is not convergence.
5. Produce one machine-readable or centrally checked V1 mandatory-issue set so future scope changes cannot leave five authorities disagreeing.
6. Re-run the integrated #106 acceptance matrix after scope-convergence changes land.

## Acceptance criteria

- A reviewer can start from #33 and derive one unambiguous list of all V1.0.0.0 mandatory authorities/work items.
[...]

Issue #101, opened 2026-09-16
[V1.0.0.0][API-FREEZE] Freeze public C++/QML/CMake surface for 1.x compatibility


skawu:
## Parent / goal

Child of #33 and #39. Before `v1.0.0.0`, freeze the application-facing HyRemote surface that will carry 1.x compatibility expectations.

This is a review/fix gate, not a request to expand V1 product scope.

> **V1 convergence authority update (2026-09-16):** all implementation/review for this gate is now absorbed into Draft PR #106, the single active V1 product line. Earlier `feature/101-v1-api-freeze` branch wording is historical. The package target inventory below is aligned with the final minimal V1 artifact contract and supersedes the original wording that listed `HyRemote::Core` as an installed SDK target.

## Frozen architecture boundary

The stable application product model is:

- Embedded C++ applications use the single normal application-facing target `HyRemote::RemoteAccess`;
- Declarative applications use `import HyRemote` / QML `RemoteAccess` over that same shared runtime;
- Transparent QPA is a separately packaged exact-Qt platform MODULE payload reached through launch/deployment, not an application link target;
- `hyremote-core` remains STATIC source/internal composition only and is not installed/exported as a V1 application SDK target;
- application developers do not assemble Core Session/CaptureSource/InputSink/Transport/backend types;
- no QPA/private Qt type enters the stable C++/QML API;
- transport choice does not change the public application contract;
- `BUILD_SHARED_LIBS` must not create alternate V1 product personalities.

Do not introduce new convenience APIs merely because V1 freeze is occurring.

### C++

- installed `HyRemote/RemoteAccess.h`;
- exported enums/structs, ownership/move semantics, error/lifecycle/configuration contracts;
- `connectedClientCount()` backend-neutral diagnostic introduced by #91;
- symbol visibility/export discipline;
- headers must not leak Core/private/backend/QPA implementation types;
- construction remains inert, loopback/5900 remain defaults, and remote input remains disabled by default.

### Declarative QML

- URI/type identity and creatable `RemoteAccess` type;
- property names/types/read-only semantics;
- state/error enums;
- lifecycle/configuration semantics relative to C++;
- initial `enabled: true` applies only after `componentComplete()` so declaration order does not become lifecycle glue;
- `connectedClientCount` parity;
- no backend/client/socket object exposure.

### CMake / installed SDK

Freeze the normal installed application contract to:

- package name `HyRemote`;
- **one normal C++ imported target: `HyRemote::RemoteAccess`**;
[...]

PR #191, opened 2026-09-19
fix(core): #164 contain exceptions at the backend-invoked callbacks

skawu:
## What this is

The first increment of **#164**, covering the **exception boundary** part of its required work and **acceptance
criterion 1**.

## The gap

`onFrameReady`, `onCaptureEvent` and `onTransportEvent` are invoked by the backend from the backend's own stack - a
capture thread, a transport runtime. Their bodies do bounded bookkeeping that can allocate (a mailbox push, a rejection
reason), and an exception from there would unwind into code HyRemote does not own. The existing input path already had
this contract, and says so in its own comment: "the exception must not unwind through the transport runtime". The three
callbacks did not.

## The change

The same boundary now wraps all four callbacks: at `session.cpp` the callbacks are invoked through
`Impl::atCallbackBoundary`, which reports through the existing error model as a recoverable `ComponentFailure` and never
lets anything through. The handler deliberately **does not allocate** - the failure it is handling may be `bad_alloc`,
so formatting a message there would throw out of a `noexcept` boundary - and it is best-effort by design, documented in
place, because the alternative is terminating the process in the caller's stack.

## Verification

`hyremote-core-test-callback-exception-boundary` (new, registered in the Core test list) injects allocation failure by
replacing the global allocation function in that binary and **sweeping** the injection position (0..63 for frames, 1..31
for capture events) rather than guessing one, so no single allocation ordering makes it pass by accident. A third case
proves the reporting half on the input path with a sink that throws: no escape, `ComponentFailure`, recoverable, session
still usable.

| Build | Result |
|---|---|
| boundary reverted (probe) | all three cases pass - see the gap below |
| change in place | three cases pass, suite **76/76** in the QML+QPA configuration, ten gates pass |

## The gap this does not close, stated plainly

The three callbacks turn out to be **allocation-free** on every path the fakes can drive: a valid frame allocates
nothing, an invalid frame's rejection reason fits in the small-string buffer, and the capture-event fault path copies
nothing that allocates. The single allocation a delivery performs is the test fake's own `std::function` handler copy.
So the sweep proves *"nothing escapes at any reachable position"* and the input case proves *"what is caught is
converted"*, but neither yet proves that **the new boundary** is what caught a failure - reverting it makes the whole
test pass, which is how the gap was found rather than assumed.

Closing it needs an injection on a path that allocates inside the callback. The `ProducerThrottle` overflow branch is
the natural candidate: its diagnostic message is far longer than the small-string buffer, and
`test_mailbox.cpp:114-122` already shows how to drive that push result. That is a follow-up, recorded here instead of
being claimed.

## Still open in #164

- the overflow-path injection above, to turn the boundary's coverage from defensive into proven;
- the remaining required work: `noexcept` correctness audit, composite input exactly-once semantics under child
  failure, bounded Quick empty-grab retry, CPU-frame stride validation, and the exported-symbol surface audit
  (criteria 2-7).

Refs #164 #101 #141 #147 #157 #159


skawu:
## Attempted the follow-up: the overflow-path injection does not work yet, and it found something else

I tried to close the gap described above by injecting on the `ProducerThrottle` overflow branch, whose diagnostic
message is far longer than the small-string buffer and therefore allocates inside `onFrameReady`. The attempt is
**not** part of this PR - the test file was restored to the committed state - but what it did is worth recording.

**The setup**: `SessionConfig{frameQueue.capacity = 1, frameQueue.policy = ProducerThrottle}`, an `EnqueueGate`
closed and installed with `FakeTransport::setGate`, one delivery to park the dispatch worker inside
`enqueueFrame()`, verified deterministically with `gate.waitForEntered(1)` (it returns true), then a few further
deliveries to fill the single waiting slot and trip the overflow branch.

**The result**: the **delivering thread hangs** in the further-delivery loop, reproducibly, before any injection is
even relevant - the test reaches `[TRACE] parked, entered=yes` and never prints the next line, with the global
allocation counter disarmed. It hangs with the boundary change in place and, as far as the run got, at position 0 of
the sweep.

**What it is not**: the dispatch worker does **not** hold `impl.mutex` across the transport call - `session.cpp`
calls `transport->enqueueFrame(...)` and only takes the lock afterwards, which is the right order. So this is not the
obvious "lock held across a blocking transport call" mistake.

**Why it is recorded rather than fixed here**: the cause is not established, and guessing at a fix for a hang on a
delivery path - one that a capture thread would be sitting in - is exactly the kind of change this repository asks to
be evidence-backed. So it stays open, with a deterministic repro above, rather than being papered over.

**Consequence for this PR**: the gap stands. The sweep proves nothing escapes at any reachable position and the input
case proves that what is caught is converted, but the boundary is still not proven to be what caught a failure.
Two things now need the same next step: whether a stalled transport may block a delivery indefinitely, and whether the
overflow branch can be reached deterministically enough to inject into it.

skawu:
## Two facts measured on the hang, so nobody repeats this path

Checked while investigating the blockage recorded above, and both narrow it:

1. **It is not a lock held across the transport call.** `session.cpp`'s `runDispatcher` pops the frame first and
   calls `transport->enqueueFrame(...)` at `:683` **without** `impl.mutex` held - the lock is taken only afterwards at
   `:691`, and the comment at `:679-680` states the intent ("No Core lock is held across the transport call"). So the
   parked worker is holding no Core lock while the delivering thread blocks.
2. **The mailbox has no separate translation unit** - there is no `src/core/src/mailbox.cpp`, so `detail::Mailbox` is
   header-implemented and its own synchronisation has not been read yet.

So the open question is now narrow and specific: **does a delivery block by design under `ProducerThrottle` once the
transport stalls?** If `push` waits for space (throttling the producer rather than rejecting) while the dispatch worker
is parked inside `enqueueFrame()`, then the caller sitting in the callback - in production, the capturing thread - waits
indefinitely. `test_mailbox.cpp:114-122` shows `push` returning `RejectedOverflow` for a full mailbox, which suggests it
does not wait at that level, so the blocking must be somewhere between the delivery and that call.

That is a hypothesis with a repro, not a finding, and it is deliberately not turned into a speculative fix. The
next step is to read the header-implemented mailbox's push/pop synchronisation and decide whether the producer may
block there at all.

skawu:
## Correction, and the hang narrows: the transport-stall scenario itself is sound

**First, a correction to my earlier comment.** I wrote that the delivery hang appeared "with the global allocation
counter disarmed". That was an inference from the trace, and it was **wrong**: the delivery loop in that version ran
*inside* the armed region (the counter was armed immediately before it), so an injected `bad_alloc` was in play. The
trace could not distinguish the two, and I should not have written the stronger claim.

**Re-measured without arming anything**, same scenario (`capacity = 1`, `ProducerThrottle`, `EnqueueGate` closed and
installed, one parking delivery verified with `waitForEntered`, then four deliveries):

```
parking frame delivered=1, entered=1
after deliver 0  delivered=1 framesRejectedOverflow=0 framesAccepted=2
after deliver 1  delivered=1 framesRejectedOverflow=1 framesAccepted=2
after deliver 2  delivered=1 framesRejectedOverflow=2 framesAccepted=2
after deliver 3  delivered=1 framesRejectedOverflow=3 framesAccepted=2
deliveries done; opening gate and stopping
stopped
```

So the scenario is **good**: the overflow branch is reached deterministically from the first extra delivery, the
mailbox stops accepting at capacity as the policy says, and nothing blocks. The case then runs to completion.

**Two candidates eliminated by reading, not by guessing**:

- it is not a lock held across the transport call (`runDispatcher` calls `transport->enqueueFrame` at
  `session.cpp:683` with no Core lock held; the lock follows at `:691`);
- it is not a blocking mailbox push: `detail::Mailbox::push` (`src/core/src/detail/mailbox.hpp`) is non-blocking by
  construction - it returns `Stored` / `StoredAfterDroppingOldest` / `RejectedOverflow` / `RejectedClosed`, and only
  `waitForWork()`, which the dispatch worker uses, ever waits on the condition variable.

**What that leaves**: the hang is **injection-induced** - the same deliveries complete when nothing is armed and did
not when a single allocation was made to fail - so an allocation failure somewhere on the delivery/dispatch path can
leave a delivery stuck. Which position, and therefore which mechanism, is not established yet; the sweep cannot say,
because the process simply stops making progress.

**Where this leaves this PR**: unchanged - the change and its evidence are as committed, and the instrumentation was
kept out of the tree (the test file is identical to the committed version). The follow-up is now sharper and cheaper:
bisect the armed position, or instrument the delivery call per step, to find the first position that stops making
progress.

**A related fact worth having while it is fresh**: with `capacity = 1` and `ProducerThrottle`, `framesAccepted` stops
at 2 - one frame owned by the dispatch worker (ADR-0003) plus one waiting - which matches the documented ownership
accounting exactly.

skawu:
## Hang bisected: it needs an injected failure on a thread other than the caller

Two controlled runs, same stall scenario (`capacity = 1`, `ProducerThrottle`, closed `EnqueueGate`, one parking
delivery verified with `waitForEntered(1)`), differing only in **which thread may consume the injection**:

**A. Injection armed per delivery on the calling thread** (four deliveries, one session): every delivery completes.
`framesRejectedOverflow` stays 0 for the positions consumed before the callback and the session stays usable - the
runs finish, nothing blocks.

**B. Injection restricted to the arming thread, loop shape that previously hung** (fresh session per position, gate
closed, four deliveries, then `open()` and `stop()`), eight positions:

```
position 0..7: arming for this thread only
position N:    deliveries done escaped=0 overflow=3 accepted=2
position N:    stopping / stopped
all positions completed
```

**So the hang requires an allocation failure on a thread other than the one driving the delivery.** With the
injection pinned to the caller, the shape that hung before completes all eight positions. That is a mechanism, not a
guess: it was the one variable changed between the hanging run and this one.

**Why that matters more than the hang itself**: the other threads here are Core's own workers, and a worker's thread
function has no top-level exception guard - `runDispatcher` (`session.cpp:655`) catches only around
`transport->enqueueFrame` (`:682-688`), while its bookkeeping afterwards (`:691-698`, which builds a `SessionError`)
is unprotected. An exception escaping a `std::thread` function terminates the process, which is exactly what #164's
acceptance criterion 2 forbids ("no path can terminate the process because of an avoidable internal
allocation/formatting exception"). The hang observed earlier is therefore a symptom of an unprotected worker path,
not of the callback boundary.

**Two concrete follow-ups, both cheap now**:

1. Ship the overflow case with the injection pinned to the calling thread (`std::atomic<std::thread::id>` compared in
   the replaced `operator new`), assert the boundary's own message appears, and revert the boundary as the
   two-direction probe. With the injection pinned this is deterministic and it finally makes the boundary proven
   rather than defensive - the gap this PR currently records as open.
2. Decide what a worker-thread exception should do. Criterion 2 says it may not terminate the process, so the
   worker needs a boundary of its own, and the failure it reports has to follow the same error model as the callback
   boundary.

**Tree state**: the instrumentation never entered the tree - `git diff` against the committed version is empty and the
working tree is clean, verified immediately before posting this.

skawu:
## Worker-thread boundary added, and the hang is not what I thought

The bisect left one conclusion standing - the hang needs an injected failure on a thread other than the caller - so
the obvious next move was to guard the worker threads and see whether the hang goes away. It does **not**, and that
is worth recording precisely.

**Added in this commit**: `Impl::atWorkerBoundary` + `Impl::reportWorkerFailure` (`src/core/src/session.cpp`), applied
at the two `std::thread` launch sites. Worker threads are `std::thread` entry points, so an escaping exception
terminates the process - forbidden by acceptance criterion 2. `runDispatcher` guarded only
`transport->enqueueFrame` (`:682-688`) while its bookkeeping afterwards (`:691-698`, which constructs a
`SessionError`) was unprotected, and `runScheduler` had no guard at all. A failing worker now reports a
non-recoverable `ComponentFailure`, so the session faults instead of staying `Running` with a dead worker, and the
thread function returns normally so the owning join stays well defined.

**The hang survived that change**, so it is not an escaping worker exception. Re-running the stall scenario with the
injection free to be consumed by any thread, instrumented per step, position 0:

```
[WORKER-PATH] position 0: deliveries begin
[WORKER-PATH] position 0: deliveries done escaped=1
```

The deliveries **complete**, with one exception escaping into the caller. That escape comes from the harness, not
from the boundary: `FakeCaptureSource::deliverInternal` copies its `std::function` handler before invoking it
(`support/fakes.hpp`), and that copy allocates. The delivery therefore fails *before* the callback is entered, so the
session never sees the completion - and what hangs is the next thing that runs, which is `~RunningSession` calling
`stop()` during unwinding of the failed check. The `[WORKER-PATH] ... opening the stall and stopping` line never
prints, which places the blockage inside teardown rather than in the delivery loop.

**Two hypotheses eliminated, both by reading rather than guessing**:

- *an exception escaping a worker thread* - the boundary above removes that possibility, and the hang remains;
- *in-flight capture requests never draining* - `impl.inFlight` is reset during teardown (`session.cpp:674` and
  `:848`), so teardown does not wait for it.

**What that leaves, stated as the next step rather than a conclusion**: `stop()` blocks when a capture completion is
lost before it reaches the session. Finding what teardown waits on that such a completion would have satisfied is the
way in, and it is a product-level hazard worth its own attention, because a `stop()` that never returns means an
application that cannot shut down. It belongs with the lifecycle-bounds work in this issue rather than in this
commit.

**This commit is therefore defensive hardening, not a proven fix**, and the test case that would have exercised it
was deliberately left out of the tree rather than shipped as a hanging test. Suite 76/76 in the QML+QPA
configuration; ten gates pass. The test file is unchanged by this commit.

skawu:
## The hang is explained - and my last two comments were wrong about why

Reading `teardownRun` (`src/core/src/session.cpp:621-681`) end to end closes this, and the answer is not what I
said before. I am correcting it here rather than leaving a wrong explanation in place.

**What teardown actually does**, in order: publish the terminal state, `gate->closeAndDrain()` (`:655`),
`stopComponentQuietly(source)` (`:657`), `mailbox->close()` (`:660`), **join both workers** (`:663-666`),
`stopComponentQuietly(transport)` (`:669`), and only then, at `:676`, `impl.teardownStarted = false;`.

**The hang, explained**: in my case the failing assertion aborted before `stall.open()`, so the dispatch worker was
still parked inside the **test fake's** `EnqueueGate::enter()`. Teardown then reached `schedulerThread.join()` /
`dispatcherThread.join()` (`:663-666`), and that join cannot return while the worker is waiting for a gate only the
test could open. The missing `opening the stall and stopping` print was the clue I should have followed: it says the
gate was never opened, which makes the blockage a **test-harness deadlock**, not a Core defect. My earlier claims -
that the hang needed an injected failure on another thread, and that it was an escaping worker exception - were both
wrong, and the pinning experiment that "proved" the first one only ever changed how quickly the case reached its own
abort path.

**Two genuine hazards this exposed anyway, and they are the reason to keep working here**:

1. **The teardown claim leaks on any exception.** `teardownStarted` is released only at `:676`, and nothing between
   `:651` and `:676` is guarded - `gate->closeAndDrain()`, `mailbox->close()`, both `join()` calls (which can throw
   `std::system_error`) and the two `notify_all()`s. If any of them throws, the claim stays set, the state stays
   `Stopping`, and **every later `stop()` blocks forever** on the claim wait at `:570`
   (`cv.wait(lock, [&impl] { return !impl.teardownStarted; })`). `stop()` is supposed to be idempotent, so this is a
   real defect and it is fixable structurally - release the claim and publish the terminal state on every path,
   rather than only on the happy one.
2. **Teardown joins workers that may be blocked inside the transport.** The comment at `:679-680` states the
   contract: "No Core lock is held across the transport call: the hand-off is a bounded post/enqueue by contract".
   Teardown relies on that bound for its join to return. A transport that blocks violates the contract, and then
   `stop()` cannot return at all - which is a product-level outcome (an application that cannot shut down), and it
   belongs with the bounded-occupancy work in #158 rather than being solved by detaching a worker that owns Core
   state.

**Why no timeout is the answer to (2)**: bounding the join with a timeout means either the join is abandoned and the
thread later touches `impl` (undefined behaviour), or the process leaks a live thread into its own teardown. So the
honest fix is (1) structural, plus a contract-and-bound decision for the transport, which the owner has already been
asked for and has now directed: **bounded teardown, in #164, with the plan presented before it lands**.

**Consequences for this PR**: nothing changes in the committed code. The worker boundary stays (criterion 2 requires
it whether or not it addresses this hang), and the case that produced the hang stays out of the tree. The next
commit will carry the claim-release fix with a deterministic test - inject a failure inside a teardown step, then
require that a second `stop()` still returns.

skawu:
## Rebased onto the merged candidate line, and the boundary proof is not achievable this way

**Rebase**: `develop` moved to `23b5edd` when **#187 (#159)** merged, and this branch touched the same file, so the
branch is now rebuilt on top of it - commits `129b152` (callback boundary) and `b63f6ed` (worker boundary). Full
suite **76/76** in the QML+QPA configuration, all **ten** gates pass, and the boundary test itself is 3/3.

**The proof attempt failed, and it is not shipped.** I tried to turn the callback boundary from defensive into
proven by injecting on the one path that allocates inside `onFrameReady` - the `ProducerThrottle` overflow branch,
whose diagnostic is far longer than the small-string buffer - with the injection **pinned to the delivering thread**
via an `atomic<std::thread::id>` compared in the replaced `operator new`, and swept **0..31**. Across every position
the boundary never reported: no swept allocation position lands inside `onFrameReady` on that path, so the injected
failure either lands in the harness's own pre-callback allocations or never reaches the callback's allocation at all.

Rather than weaken the assertion into something that passes without proving anything, the case was dropped: the
branch carries the **same test file as before**, and this PR's body keeps stating that the boundary is defensive
rather than proven. That gap is real and it is recorded here instead of being papered over.

**What would actually close it**, for whoever picks this up: a deterministic seam *inside* the callback - for example
a statistics or fault-reporting hook that a test can make fail - or an internal allocation the test can force on that
path. Until then, the honest statement is the one in the body.

**State**: branch rebased and pushed; suite and gates green; the test file unchanged by this rebase.

skawu:
V1 pre-merge review: this PR is **not currently eligible to merge** onto the security convergence line.

Reasons measured on the exact current PR head `b3a63f42e61dd2ae9789eb23c77c835d22db74a1`:

1. GitHub reports the PR non-mergeable against current `develop`; its base predates the accepted #192/#193 security work.
2. Exact-head Hosted evidence is not green: Quick, RemoteAccess, QML, V1 GA, Widgets and Bounded RFB workflows are failed; only Git Flow, SDK consumption and QPA are green.
3. The PR description itself records that the allocation sweep does not yet prove the new callback boundary is what catches the failure. #164 acceptance cannot treat a defensive boundary as proven merely because the surrounding test passes when the boundary is reverted.

The implementation direction is still valid: backend callback/worker exception containment and composite-input exactly-once semantics belong to #164. Before merge, topology-merge current `develop`, resolve any overlap, add the deterministic overflow-path injection identified in the PR, and obtain a fresh exact-head nine-workflow envelope. Do not weaken tests or security/build changes to recover green.

No product architecture change is requested. Refs #164 #159 #143 #170

